PR #69, opened 2019-06-24
feat(accordion): implement accordion with core-toggle

### LINKED ISSUE #59: Component: Accordion
- [x] Initial design
- [x] Initial package

piofinn:
affects: @fremtind/jkl-accordion

## 📥 Proposed changes

Move implementation of Accordion to use `@nrk/core-toggle` to ensure good accessibility features. Also makes use of the chevron ornament mixin from #62.

## ☑️ Submission checklist

## 💬 Further comments

Closes #59

`core-toggle` needs to be polyfilled for IE. This should maybe be done by consumers on a per project-basis?


packages/accordion-react/package.json
```diff
@@ -25,7 +25,8 @@
     },
     "dependencies": {
         "@fremtind/jkl-accordion": "^0.0.1",
-        "@fremtind/jkl-utils": "^0.0.1"
+        "@fremtind/jkl-utils": "^0.0.1",
```
lfbergee: Not in use?

packages/accordion-react/src/Accordion.tsx
```diff
@@ -1,26 +1,33 @@
 import React, { ReactNode, useState } from "react";
+// @ts-ignore
```
lfbergee: Maybe a comment to say why, we should probably not encourage ts-ignore all willy nilly (https://www.merriam-webster.com/dictionary/willy-nilly)

packages/accordion-react/src/Accordion.tsx
```diff
@@ -1,26 +1,33 @@
...
+import CoreToggle from "@nrk/core-toggle/jsx";
 
 interface Props {
     children: ReactNode;
 }
 
-export const Accordion = ({ children }: Props) => <div className="jkl-accordion">{children}</div>;
+export function Accordion({ children }: Props) {
```
lfbergee: Split into two files? Then both can be Props and its maybe a bit eaier to read?

packages/accordion-react/src/Accordion.tsx
```diff
@@ -1,26 +1,33 @@
...
-            <button className="jkl-accordion__item-title" onClick={() => setIsOpen(!isOpen)} tabIndex={0}>
-                <div className="jkl-accordion__item-title-text">{title}</div>
-                <div className="jkl-accordion__item-title-icon" />
+        <div className={`jkl-accordion-item${openClassName}`}>
+            <button className="jkl-accordion-item__title">
+                <div className="jkl-accordion-item__title-text">{title}</div>
+                <div className="jkl-accordion-item__title-icon" />
             </button>
-            <div className="jkl-accordion__item-content">{children}</div>
+            <CoreToggle hidden onToggle={() => setIsOpen(!isOpen)}>
```
lfbergee: named function please

packages/accordion-react/src/Accordion.tsx
lfbergee: And we should consider adding types to the upstream NRK project if we keep using this.

packages/accordion/accordion.scss
```diff
@@ -1,23 +1,22 @@
...
-                transform: rotate(135deg);
-            }
-            & .jkl-accordion__item-content {
-                height: auto;
-            }
+.jkl-accordion-item {
+    margin-bottom: 2.5em;
+    &--expanded {
+        & .jkl-accordion-item__title-icon {
+            @include chevron($chevron-size, $chevron-color, $weight: $chevron-weight, $state: open, $rotate: 180);
```
lfbergee: I like this

packages/accordion-react/src/Accordion.tsx
piofinn: Agreed. Added comment about this

packages/accordion/accordion.scss
piofinn: Thx!

packages/accordion-react/src/Accordion.tsx
piofinn: - [x] Done!

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-    - "lts/dubnium"
+    - "--lts"
```
lfbergee: It seems like this might have been connected to the Verizon outage yesterday. Should work with both --lts, lts/dubnium, 12 and lts/* now.

packages/accordion-react/src/AccordionItem.tsx
```diff
@@ -0,0 +1,28 @@
...
+    children: ReactNode;
+    startExpanded?: boolean;
+}
+
+export function AccordionItem({ children, title, startExpanded = false }: Props) {
+    const [isOpen, setIsOpen] = useState(startExpanded);
+    const openClassName = isOpen ? " jkl-accordion-item--expanded" : "";
+    const onToggle = () => setIsOpen(!isOpen);
+    return (
+        <div className={`jkl-accordion-item${openClassName}`}>
```
lfbergee: Should we allow users to add their own class to extend the styling?

packages/accordion-react/src/AccordionItem.tsx
```diff
@@ -0,0 +1,28 @@
...
+            <button className="jkl-accordion-item__title">
```
lfbergee: What does the button actually do here? I might be misunderstanding this, but there is no eventHandler on the button, so should this not be a div or span or something other than a button?

piofinn: `type="button"` to the rescue!

PR #70, opened 2019-06-24
feat(logo): add logo package, better animate logo

### LINKED ISSUE #38: Split logo-react into logo and logo-react package
- [x] Logo style into its own package
- [x] Fix animated logo animate properly

lfbergee:
affects: @fremtind/jkl-header, @fremtind/jkl-logo-react, @fremtind/jkl-logo

## 📥 Proposed changes

Add logo package and react wrapper

## ☑️ Submission checklist

## 💬 Further comments

Closes #38


lfbergee:
@all-contributors please add @Steinop for design

packages/logo/logo.scss
```diff
@@ -0,0 +1,50 @@
+@import "~@fremtind/jkl-core/build/scss/variables/colors.scss";
+
+$jkl-logo-color: $svart;
+$jkl-logo-color--negative: $snøhvit;
+$jkl-logo-magic-origin--f: 30px 132px; /* Is connected to the viewbox of the logo, change at your own risk */
+$jkl-logo-magic-origin--line: 34.3px; /* Must be the same as the line x1 origin, change at your own risk */
+
+.jkl-logo {
+    &--animated {
```
Saegrov: I think the current animation feels a bit weird. There are some weird bouncing going on and also the animation is quite different from the example on brandspace.

packages/logo/logo.scss
```diff
@@ -0,0 +1,50 @@
...
+                    opacity: 0;
+                    transform: translateX(10%);
+                }
+            }
+        }
+    }
+}
+
+.jkl-logo__wrapper {
+    transition-delay: 0ms;
```
Saegrov: Do we need transition-delay: 0 ?

packages/logo/example/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+import ReactDOM from "react-dom";
+import "../logo.scss";
+
+interface Props {
+    negative: boolean;
+    kicked: boolean;
+}
+
+const Logo = ({ negative, kicked }: Props) => (
+    <svg
```
Saegrov: Is there a way we can use logo-react so that we don't have to maintain 2 separate logos? Now after my update it is out of sync and this will probably happen in the future as well.

packages/logo/logo.scss
piofinn: We have to look at core animation soon regardless, so this is a great place to look for inspiration for the more expressive animations. I think some "deep" s-curves, `ease-in-out`-style could look good here on most transitions. Maybe even some linears 😮

packages/logo/example/index.tsx
piofinn: That would make a circular dependency 😕

packages/logo-react/src/LogoAnimated.tsx
```diff
@@ -1,31 +1,65 @@
...
+                className={`jkl-logo jkl-logo--animated ${
+                    showSymbol ? "jkl-logo--animated__symbol-only" : ""
+                } ${className}`}
+                width={width}
+                viewBox="30 18 658 127"
+                xmlns="http://www.w3.org/2000/svg"
+                fill="currentColor"
+                stroke="currentColor"
+            >
+                <g>
```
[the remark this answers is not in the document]
lfbergee: Yes, I should run the SVG optimizer on it, it might collapse some of the lines into one as well. Will do tomorrow.

lfbergee:
I clicked resolve conversation instead of add comment, sorry. 

Yes, we could put the SVG in the style package and depend on that, adding the SVG bundler plugin to roll up.

The transform delay 0 is needed for the animation to only have a delay on the forward motion of the animation, not the return animation.

packages/logo/logo.scss
lfbergee: Yeah, let's find some good standardized curves to work with, there is too much custom stuff already

packages/logo-react/src/LogoAnimated.tsx
```diff
@@ -1,31 +1,65 @@
...
+                    <g className="jkl-logo--animated__F">
+                        <line
+                            className="jkl-logo--animated__F-line"
+                            y2="32.563024"
+                            x2="99.846844"
+                            y1="32.563024"
+                            x1="34.3"
+                            strokeWidth="8"
+                            fill="none"
```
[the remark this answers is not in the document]
lfbergee: SVG optimizer should take care of that

packages/logo-react/src/LogoAnimated.tsx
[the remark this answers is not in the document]
lfbergee: SVG optimized

packages/logo/logo.scss
lfbergee: Reverted to more basic animation curves. We should still standarize these. Do that in a own PR.

PR #81, opened 2019-06-25
feat(checkbox): add checkbox

Saegrov:
## 📥 Proposed changes

New checkbox

## ☑️ Submission checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._


packages/core/src/functions.scss
```diff
@@ -0,0 +1,9 @@
+
+/**
+ * Function for converting a px size to rem;
+ */
+@function rem($size) {
```
piofinn: Maybe rename `$size` to `$pxSize`? A bit nitpicky, but since unitless values don't work any clarification is good (it shows up in intellisense) 😊

packages/checkbox/checkbox.scss
```diff
@@ -0,0 +1,89 @@
+@import "~@fremtind/jkl-core/build/scss/variables/colors";
+@import "~@fremtind/jkl-core/build/scss/functions";
+
+$checkbox-size: 1.5rem;
```
piofinn: There are a few other values that would be nice to parametrize as well (colors, border-widths)

packages/checkbox/checkbox.scss
```diff
@@ -0,0 +1,89 @@
...
+        border: solid rem(1px) $svart;
+        height: $checkbox-size;
+        margin-right: 0.5rem;
+        position: relative;
+        transition: background-color 200ms ease, transform 150ms ease;
+        vertical-align: middle;
+        width: $checkbox-size;
+
+        &:focus {
+            outline-offset: rem(2px);
```
piofinn: Not supported in IE ☹️

packages/checkbox/checkbox.scss
```diff
@@ -0,0 +1,89 @@
...
+    cursor: pointer;
+    display: inline-flex;
+    padding: rem(5px);
+
+    &__input {
+        -webkit-appearance: none;
+        background-color: $snøhvit;
+        border: solid rem(1px) $svart;
+        height: $checkbox-size;
+        margin-right: 0.5rem;
```
piofinn: Remember to use the spacing scale. `$component-spacing--small` is the equivalent here

Saegrov: I thought the component scale was mainly for vertical spacing - not horizontal?

packages/core/src/functions.scss
Saegrov: Not nitpicky at all - I agree. Fixed.

packages/checkbox/checkbox.scss
Saegrov: Aye. Fixed.

packages/checkbox/checkbox.scss
Saegrov: hello

packages/checkbox/checkbox.scss
piofinn: I'm thinking that the layout scale is for vertical, but the component spacing is for both? Sounds like we should have a chat with design about this one..

PR #92, opened 2019-06-27
feat(bullet list)

### LINKED ISSUE #45: Components: BulletList
- [x] Initial design
- [x] Initial package

lfbergee:
## 📥 Proposed changes

Closes #45

## ☑️ Submission checklist


packages/bullet-list/example/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+import ReactDOM from "react-dom";
+import "../bullet-list.scss";
+
+const App = () => (
+    <>
+        <ul className="jkl-bullet-list">
+            <li className="jkl-bullet-list__item">Oh hi Mark</li>
+            <li className="jkl-bullet-list__item">Oh hi Johnny</li>
+            <li className="jkl-bullet-list__item">How is your life</li>
+            <li className="jkl-bullet-list__item--nested">Oh, yeah, nice</li>
```
piofinn: This is not semantically correct. We should try to find a way to nest properly. Maybe something like:
```scss
.jkl-bullet-list {
    & > &__item {
        // main style
    }
    & & > &__item {
        // nested style
    }
    & & & > &__item {
        // even more nested style
    }
}
```
Then we could do normal nesting like so
```html
<ul class="jkl-bullet-list">
    <li class="jkl-bullet-list__item">something
        <ul class="jkl-bullet-list">
            <li class="jkl-bullet-list__item">something else</li>
        </ul>
    </li>
</ul>
```

packages/bullet-list-react/src/Li.tsx
```diff
@@ -0,0 +1,10 @@
+import React, { ReactNode } from "react";
+
+interface Props {
+    children: ReactNode;
+    nested?: boolean;
```
piofinn: See comment below for proposed change. We want to remain semantically correct when nesting list elements.

piofinn: See PR #95 for proposed implementation

packages/bullet-list/example/index.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<html style="height: 100%">
+    <head>
+        <meta charset="utf-8" />
+        <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+        <title>Bullet list</title>
+        <meta name="viewport" content="width=device-width, initial-scale=1" />
+    </head>
+
+    <body style="height:100%;display:flex;align-items:center;justify-content:center;">
```
Saegrov: Opionion: I think it is better to put all styles and style imports in a `index.scss` like in button-example.

Saegrov: Now it is some inline in html and some style imports in the react component.

packages/bullet-list/example/index.tsx
Saegrov: Since you've implemented this change the example has to be updated.

packages/bullet-list/example/index.tsx
```diff
@@ -0,0 +1,22 @@
+import React from "react";
+import ReactDOM from "react-dom";
+import "../bullet-list.scss";
+
+const App = () => (
+    <>
+        <ul className="jkl-bullet-list">
```
Saegrov: There is no need to use typescript here. Use regular HTML instead.

PR #97, opened 2019-06-28
fix(build): fix deterministic build order

Saegrov:
## 📥 Proposed changes

Fix deterministic build order
Move StoryTemplate to root since it will be removed

## ☑️ Submission checklist

## 💬 Further comments


piofinn:
Looks like Travis only builds two packages at a time, so build is still kinda slow 😕Much faster than before locally, though!

PR #99, opened 2019-07-01
Implement detection of keyboard navigation

piofinn:
## 📥 Proposed changes

Implement detection of whether the user is navigating with the mouse or keyboard, and signify this with a data-attribute on the html element.

## ☑️ Submission checklist

## 💬 Further comments

I've placed this in utils, but it could presumably also be in core.


packages/utils/src/tabListener.ts
```diff
@@ -0,0 +1,30 @@
+const navKeys = ["Tab", "ArrowUp", "ArrowRight", "ArrowDown", "ArrowLeft"];
```
Saegrov: Do we alle the arrow-keys? I don't think you can use arrow keys to change focused item?

packages/utils/src/tabListener.ts
```diff
@@ -0,0 +1,30 @@
...
+let mousenavigation = false;
+
+function handleMouseDown() {
+    if (!mousenavigation) {
+        mousenavigation = true;
+        const htmlElement = document.querySelector("html");
+        htmlElement && htmlElement.setAttribute("data-mousenavigation", "true");
+        window.removeEventListener("mousedown", handleMouseDown);
+        window.addEventListener("keydown", handleKeydown);
```
Saegrov: Use document to add event listeners:
https://stackoverflow.com/a/2671217

packages/utils/src/tabListener.ts
```diff
@@ -0,0 +1,30 @@
...
+            mousenavigation = false;
+            const htmlElement = document.querySelector("html");
+            htmlElement && htmlElement.removeAttribute("data-mousenavigation");
+            window.removeEventListener("keydown", handleKeydown);
+            window.addEventListener("mousedown", handleMouseDown);
+        }
+    }
+}
+
+export function initTabListener() {
```
Saegrov: I wrote a test that you might consider adding:
```typescript jsx
import { fireEvent } from "@testing-library/dom";
import { initTabListener } from "./tabListener";

let mouseNavigation = "data-mousenavigation";

test("should initial mouseNavigation set to null before clicks or tabs", () => {
    initTabListener();

    const htmlElement = document.querySelector("html");

    if (!htmlElement) {
        throw 'Something went wrong  - couldn\'t find body element';
    }

    expect(htmlElement.getAttribute(mouseNavigation)).toBe(null);
});

test("should add data-mousenavigation attributen when clicking a button", () => {
    initTabListener();

    const bodyElement = document.querySelector("body");
    const htmlElement = document.querySelector("html");

    if (!bodyElement || !htmlElement) {
        throw 'Something went wrong  - couldn\'t find body element';
    }

    const buttonElement = document.createElement("button");
    bodyElement.appendChild(buttonElement);

    fireEvent(buttonElement, new MouseEvent("mousedown", { bubbles: true }));

    expect(htmlElement.getAttribute(mouseNavigation)).toBe("true");
});

test("should remove data-mousenavigation attribute when tabbing after a click", () => {
    initTabListener();

    const bodyElement = document.querySelector("body");
    const htmlElement = document.querySelector("html");

    if (!bodyElement || !htmlElement) {
        throw 'Something went wrong  - couldn\'t find body element';
    }

    const buttonElement = document.createElement("button");
    bodyElement.appendChild(buttonElement);

    fireEvent(buttonElement, new MouseEvent("mousedown", { bubbles: true }));

    expect(htmlElement.getAttribute(mouseNavigation)).toBe("true");

    fireEvent(document, new KeyboardEvent("keydown", { key: "Tab", bubbles: true }));

    expect(htmlElement.getAttribute(mouseNavigation)).toBe(null);
});

```

packages/utils/src/index.ts
```diff
@@ -1,3 +1,4 @@
 import { StoryTemplate } from "./internal/StoryTemplate";
+import { initTabListener } from "./tabListener";
```
Saegrov: IMO this should be part of core and not util.

packages/utils/src/tabListener.ts
Saegrov: Oh. And we need to install @testing-library/dom if we use this test.

packages/utils/src/tabListener.ts
piofinn: They are used to navigate radio buttons, tabs, dropdowns and a few other structures: https://webaim.org/techniques/keyboard/#testing

packages/utils/src/index.ts
piofinn: Moved over now, but we should refactor the entire `core` package sometime soon.

packages/utils/src/tabListener.ts
piofinn: Nice tests! They are now shamelessly stolen 😅

PR #128, opened 2019-07-05
feat(dropdown): refactor and redesign dropdown

### LINKED ISSUE #105: Dropdown improvement
New styling for dropdowns. Designs in Figma:
https://www.figma.com/file/TkbB9ANfejDSjB2u4u1OEuqM/J%C3%B8kul-components?node-id=262%3A0

### LINKED ISSUE #122: Dropdown implementation is not accessible
**Describe the bug**
When using a VoiceOver in chrome/safari it is not possible to access dropdown menu options.
When tab navigating, there is no focused state indicating that you can open the dropdown menu.
Normal dropdown tab navigation doesn't work.

Saegrov:
fixes #122
fixes #105

## 📥 Proposed changes

Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.

## ☑️ Submission checklist

-   [ ] I have read the [CONTRIBUTING]() doc

## 💬 Further comments


Saegrov:
Currently WIP. Todo-list:
- Make sure all css-props are necessary - many might be just some previous experimentation
- Have a system for z-index. This component needs z-index, but we need a system for it (in jkl-core?)
- The dropdown har margin between the option elements so there is "dead-space" between each option where no selection is possible
- The option elements are not 100% width which creates even more "dead-space" where you cant select an option
- Maybe we should make core-toggle a peer-dependency instead of a dependency?
- Right now, the components doesn't have a bottom margin. In Figma there is a bottom margin.

piofinn:
> * [ ]  Have a system for z-index. This component needs z-index, but we need a system for it (in jkl-core?)

Stumbled across Carbon's implementation of this in an example:
```scss
$z-indexes: (
    modal : 9000,
    overlay : 8000,
    dropdown : 7000,
    header : 6000,
    footer : 5000,
    hidden : - 1,
    overflowHidden: - 1,
    floating: 10000
  );
```
May be a good starting point?

piofinn:
## Two major issues

IMO, both of these have to be worked around for this component to be considered usable

❌ This implementation breaks expected keyboard navigation: Navigation within the dropdown is _only_ possible with tab/shift + tab. Using the arrow keys (the normal way of navigating select) moves the page (or the menu if it scrolls).
❌ It is not possible to jump to values by typing. Even when the input is not supposed to be filterable, this is expected behavior.

lfbergee:
I think this is a good change, downshift had a pretty complicated API. It should be possible to adjust that to be accesible(its what is made for), but its consistent and nice to use NRK core.

I think it should be a dependency, not peer, as there is a pretty small footprint. And I assume that most project at Fremtind do not have NRK core as a dependecy today. So making it a peer depdepency only increases the work load for developers implementing Jøkul, without a tangable benefit.

zindex scale is a good idea. Lets try to keep it simple.

piofinn:
After a way too long battle the dropdown now behaves as it should, via a hook based on W3C's [dropdown example](https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html) 🎉

The hook accepts a `boolean` to enable/disable typeahead search, so it should be ready for when we implement a filterable dropdown. It should also be reusable for other components, but unfortunately it relies on the structure of `button`s within `li`s, so there is a potential for improvement here.

packages/dropdown-react/src/Dropdown.tsx
```diff
@@ -1,71 +1,115 @@
...
 }
 
-export const Dropdown = ({ label, items, className = "", initialInputValue, onChange }: Props) => {
-    const onSelectionChange = (selection: string) => onChange(selection);
+interface CoreToggleSelectEvent {
+    detail: { textContent: string };
+    target: { hidden: boolean; button: HTMLButtonElement; value: { textContent: string } };
+}
+
+function lower(value: string | undefined) {
```
lfbergee: Du you need to call toString(), with typeguard and nul check first?

Or you could 
```
const lower = (str="") => str.toLowerCase().replace(/\s/g,"")
```

packages/dropdown-react/src/Dropdown.tsx
```diff
@@ -1,71 +1,115 @@
...
+    } else {
+        focusedItem = listEl.querySelector('[role="option"]');
+    }
+    focusedItem && focusedItem.focus();
+}
+
+export const Dropdown = ({ items, initialInputValue, label, onChange }: Props) => {
+    const [selectedValue, setSelectedValue] = useState(initialInputValue);
+
+    useEffect(() => {
```
lfbergee: Does this actually do something? SelectedValue is set to initialInput in contruction. 

So if this actually should do something, initialInputValue is probably not the correct name

packages/dropdown-react/src/Dropdown.tsx
```diff
@@ -1,71 +1,115 @@
...
+        setSelectedValue(initialInputValue);
+    }, [initialInputValue]);
+
+    const [listId] = useState("dropdown".concat(uuid().slice(-8)));
+    const [hidden, setHidden] = useState(true);
```
lfbergee: Not a huge fan of hidden as the variable name, i see that its derrived from the hidden properti of coreToggle, but maybe openDropdown or something is easier to understand.

packages/dropdown-react/src/Dropdown.tsx
```diff
@@ -1,71 +1,115 @@
...
-                </div>
-            )}
-        </Downshift>
-        /* eslint-enable jsx-a11y/label-has-associated-control, react/jsx-key */
+        <div className={className}>
+            <span className={`jkl-dropdown__label ${hasSelectedValue ? "jkl-dropdown__label--hover" : ""}`}>
+                {label}
+            </span>
+            <button className="jkl-dropdown__value" aria-haspopup="listbox">
+                {hasSelectedValue ? selectedValue : "Velg"}
```
lfbergee: "Velg" should be a prop so we can avoid supporting i18n

packages/dropdown-react/src/Dropdown.tsx
```diff
@@ -1,71 +1,115 @@
...
+            </button>
+            <CoreToggle
+                id={listId}
+                role="listbox"
+                className="jkl-dropdown__core-toggle"
+                popup={label}
+                hidden={hidden}
+                onToggle={onToggle}
+                onToggleSelect={(e: CoreToggleSelectEvent) => {
```
lfbergee: Should be a named function

packages/dropdown-react/src/Dropdown.tsx
```diff
@@ -1,71 +1,115 @@
...
+                    e.target.button.focus();
+                    e.target.value = e.detail;
+                    setSelectedValue(e.detail.textContent);
+                    onChange && onChange(e.detail.textContent);
+                }}
+                aria-activedescendant={hasSelectedValue && `${listId}__${lower(selectedValue)}`}
+            >
+                <ul className="jkl-dropdown__option-wrapper" tabIndex={-1} ref={listRef}>
+                    {items.map((item) => {
+                        return (
```
lfbergee: Unnessesary return

packages/dropdown-react/src/useListNavigation.ts
```diff
@@ -0,0 +1,130 @@
...
+
+type Timer = NodeJS.Timeout | undefined;
+type SearchTerm = { term: string } | undefined;
+type Direction = "prev" | "next" | "first" | "last";
+
+export function useListNavigation(enableSearch: boolean): RefObject<HTMLUListElement> {
+    const listRef = useRef<HTMLUListElement>(null);
+    let search: SearchTerm;
+    let searchResetTimer: Timer;
+    if (enableSearch) search = { term: "" };
```
lfbergee: Please add {}

packages/dropdown-react/src/useListNavigation.ts
```diff
@@ -0,0 +1,130 @@
...
+    useEffect(() => {
+        const listElement = listRef.current;
+        if (listElement) {
+            listElement.addEventListener("keydown", (event) =>
+                handleListKeyNav(listElement, event, search, searchResetTimer),
+            );
+        }
+    }, []);
```
lfbergee: The effect should probably return a function that tears down the eventListners

packages/dropdown-react/src/useListNavigation.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    return listRef;
+}
+
+function handleListKeyNav(list: HTMLUListElement, event: KeyboardEvent, search: SearchTerm, searchResetTimer: Timer) {
+    const { key, target } = event;
+    const currentFocus = target as HTMLButtonElement;
+    switch (key) {
+        case "ArrowUp":
```
lfbergee: Maybe just me, but i truly dislike fallthrough in switchcase

lfbergee:
```
const move = (action, e, list, currentFocus) => {
  e.preventDefault();
  moveFocus(list, currentFocus, action);
}

switch(key) {
  case "ArrowUp":
    return move("prev",e,list,currentFocus);
  case "PageUp":
    return move("prev",e,list,currentFocus);
...
}
```

packages/dropdown-react/src/useListNavigation.ts
```diff
@@ -0,0 +1,130 @@
...
+                    searchResult.focus();
+                }
+            }
+            break;
+    }
+}
+
+function moveFocus(list: HTMLUListElement, current: HTMLButtonElement, dir: Direction) {
+    const thisLI = current.parentElement;
+    switch (dir) {
```
lfbergee: dir? how about action? or direction? dir is directory in my mind :stuck_out_tongue:

packages/dropdown-react/src/useListNavigation.ts
```diff
@@ -0,0 +1,130 @@
...
+
+function findItem(list: HTMLUListElement, key: string, search: SearchTerm, timer: Timer): HTMLButtonElement | null {
+    const listItems = list.querySelectorAll(`[role="option"]`);
+    if (!listItems.length) return null;
+
+    if (search) {
+        search.term = search.term.concat(key);
+        resetWhenIdle(search, timer);
+
+        for (var n = 0; n < listItems.length; n++) {
```
lfbergee: Wow! Its a for-loop :1st_place_medal: Cant remeber the last time I saw that :stuck_out_tongue:  

I dont mind the for loop, it might be more readable with a map for for each or for in, but its fine, but should probably avoid var as its hoisted.

packages/dropdown-react/src/useListNavigation.ts
```diff
@@ -0,0 +1,130 @@
...
+    return null;
+}
+
+function resetWhenIdle(search: SearchTerm, timer: Timer) {
+    if (timer) {
+        clearTimeout(timer);
+        timer = undefined;
+    }
+    timer = setTimeout(
+        () => {
```
lfbergee: Doesnt eslint complain here? Arrow functions normally expects a return value

packages/dropdown/dropdown.scss
```diff
@@ -1,132 +1,158 @@
...
+$arrow-size: $value-height;
+
+$side-padding: $component-spacing--small;
+$bottom-padding: $component-spacing--small;
+$bottom-margin: $component-spacing--large;
+
+@mixin nojump-bold {
+    // Make font bolder without increasing line length
+    font-weight: bold;
+    letter-spacing: -0.009em;
```
lfbergee: https://gph.is/2lCRJnA

packages/dropdown/dropdown.scss
```diff
@@ -1,132 +1,158 @@
...
-        border-width: 0.2rem 0.2rem 0 0;
-        display: inline-block;
-        height: 0.45rem;
-        transform: rotate(135deg);
-        vertical-align: top;
-        width: 0.45rem;
-        transition: transform 125ms ease-in;
+        &--hover {
+            // STATE SHOULD BE RENAMED
+            color: #696666;
```
lfbergee: Should be a variable?

packages/dropdown-react/src/useListNavigation.ts
piofinn: I know! You can't use map on node lists (believe me, I tried) ☹️

piofinn: I'll fix the var! I think it came from a snippet 🤔

packages/dropdown-react/src/Dropdown.tsx
piofinn: The component accepts numbers as values, so they have to be converted to strings, unfortunately. Your suggestion is a bit more elegant, though!

packages/dropdown-react/src/Dropdown.tsx
piofinn: This is Simen's work, but I think this is because of the web component loading on first render, and changing `selectedValue` back to empty/`undefined`. Therefore, the `selectedValue` is updated again after rendering the first time, to ensure the correct value is displayed to the user.

packages/dropdown-react/src/useListNavigation.ts
piofinn: Apparently not?

packages/dropdown-react/example/index.scss
```diff
@@ -0,0 +1,15 @@
+@import "~@fremtind/jkl-core/core.scss";
+@import "~@fremtind/jkl-dropdown/dropdown.scss";
```
lfbergee: Load .css instead to get with vendorprefix and minification

packages/dropdown-react/src/Dropdown.tsx
```diff
@@ -1,71 +1,114 @@
-import React from "react";
-import Downshift from "downshift";
+// @ts-ignore
+import CoreToggle from "@nrk/core-toggle/jsx";
+import React, { useState } from "react";
+import nanoid from "nanoid";
```
lfbergee: We should make a task to replace all uuids with nanoid

piofinn: Done!

packages/dropdown-react/package.json
```diff
@@ -34,7 +34,9 @@
     },
     "dependencies": {
         "@fremtind/jkl-dropdown": "^0.6.1",
-        "downshift": "^3.2.10"
+        "@nrk/core-toggle": "^3.0.4",
+        "@types/nanoid": "^2.0.0",
```
lfbergee: dev dependecy?

Issue #39, opened 2019-06-21
Components: Messagebox


lfbergee:
- [x] Initial design
- [x] Initial package

Steinop:
Figma: https://www.figma.com/file/TkbB9ANfejDSjB2u4u1OEuqM/J%C3%B8kul-components?node-id=379%3A0

PR #130, opened 2019-07-09
feat(messagebox): add error,info and success message

lfbergee:
affects: @fremtind/jkl-message-box-react, @fremtind/jkl-message-box

## 📥 Proposed changes

style and react wrapper for messageboxes

## ☑️ Submission checklist

## 💬 Further comments

closes: #39


packages/message-box-react/src/MessageBox.tsx
```diff
@@ -0,0 +1,24 @@
...
+    title: string;
+    children: ReactNode;
+}
+
+type messageTypes = "info" | "error" | "success";
+
+function messageFactory(messageType: messageTypes) {
+    return function messageBox({ title, children }: Props) {
+        return (
+            <div className={`jkl-message-box__${messageType}`}>
```
piofinn: Not BEM, should also have a "plain" `jkl-message-box`

packages/message-box/message-box.scss
```diff
@@ -0,0 +1,45 @@
...
+    min-height: $layout-spacing--xl;
+    padding: $component-spacing--xl;
+}
+
+.jkl-message-box {
+    &__title {
+        margin-bottom: $component-spacing--large;
+    }
+    &__info {
+        @include box($snøhvit);
```
piofinn: This generates some redundant rules: Only `background-color` changes between the variants, but all properties are repeated.

packages/message-box/message-box.scss
```diff
@@ -0,0 +1,45 @@
...
+        &::before {
+            @include colored-line($himmel);
```
piofinn: Same as above

piofinn:
None of the above are _super_ important if we just want to get this in before the core refactor, but they should be addressed at some point (maybe make an issue if merged as is)

packages/message-box-react/README.md
```diff
@@ -0,0 +1,11 @@
+# `message-box-react`
```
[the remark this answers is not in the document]
lfbergee: Added in style package

packages/message-box-react/src/MessageBox.tsx
lfbergee: Bemified it

packages/message-box/message-box.scss
lfbergee: fixed

Issue #46, opened 2019-06-21
Components: Datepicker


lfbergee:
- [ ] Initial design
- [ ] Initial package

PR #132, opened 2019-07-10
Datepicker 0

lfbergee:
## 📥 Proposed changes

Add initial package and design for datepicker

## ☑️ Submission checklist

## 💬 Further comments

Notes:
- Pushes content down when open, not over other content, don't know if we want that
- Uses fafafa as background to hide input field background, its using fff with fff as background in inputfield, however that do not exsist in input yet. Should we add?
- dropdown not implemented as API might change a lot with open PR

closes: #46


lfbergee:
Screenshot from 2019-07-10 15-53-52
Screenshot from 2019-07-10 15-54-09

lfbergee:
I'm having some issues writing tests that actually change the date. getBytText() seems to create an eternal loop that crashes the js-runtime.

piofinn:
I think there are some issues with testing web components out of the box with jest/testing library. @Saegrov probably knows more about this, though.

packages/datepicker/datepicker.scss
```diff
@@ -0,0 +1,70 @@
+@import "~@fremtind/jkl-core/build/scss/variables/spacing.scss";
+@import "~@fremtind/jkl-core/build/scss/variables/colors.scss";
+@import "~@fremtind/jkl-core/build/scss/typography/paragraphs.scss";
+@import "~@fremtind/jkl-core/build/scss/functions.scss";
+@import "~@fremtind/jkl-text-input/text-input.scss";
```
piofinn: We should avoid importing other stylesheets directly, as this will duplicate the rules (presuming `text-input` is used elsewhere).

packages/datepicker/datepicker.scss
```diff
@@ -0,0 +1,70 @@
...
+            &--year {
+                max-width: 30%;
+            }
+            &--month {
+                flex-grow: 2;
+            }
+        }
+
+        caption {
+            @extend .jkl-p;
```
piofinn: As above. This will include the style rules for paragraphs in duplicate for this component (since we import `paragraph.scss`). Optimally we would just give the element the `.jkl-p` class, but I guess that is not an option here. I still think we should try to get rid of the duplicate. Maybe make a paragraph `@mixin` instead?

PR #137, opened 2019-07-10
chore(travis): build in stages, only build portal on master

lfbergee:
Just a test


lfbergee:
Well, this seems like it was a pretty bad idea, as it will boot a new VM for each stage. Could perhaps be parallelized, but either way its slower than without stages.

PR #140, opened 2019-07-11
feat(core): flatten structure of core package

### LINKED ISSUE #139: Refactor `core` package
## Main issues

Currently, the `core` package is a mess, and is in great need of a refactor. I've outlined some main issues:

- [x] The file structure does not make sense. Things are not where you would think
- [x] We output several empty `.css`-files on build
- [x] Several rules are written multiple times in the final `core.css` file

## Suggestions

Here are some suggested steps to alleviate these issues:

- [x] Move to a flat structure (as already discussed) for _some of_ the files. Specifically, move all files that will have content when parsed on build, to root level
- [x] Let all "pure `scss`" files reside in folders and not get built. This would be most files in the `variables` folder, along with mixins and functions
- [x] Clean up the `@import` chain in all files to remove duplicate code exports
- [ ] Export custom properties for the variables

piofinn:
affects: @fremtind/jkl-accordion, @fremtind/jkl-bullet-list, @fremtind/jkl-button,
@fremtind/jkl-card, @fremtind/jkl-checkbox, @fremtind/jkl-core, @fremtind/jkl-divider-line,
@fremtind/jkl-dropdown, @fremtind/jkl-footer, @fremtind/jkl-grid, @fremtind/jkl-header,
@fremtind/jkl-logo, @fremtind/jkl-radio-button, @fremtind/jkl-text-input

## 📥 Proposed changes

Flatten structure of `core` package as proposed in issue #139 

## ☑️ Submission checklist

## 💬 Further comments

Fixes #139


packages/bullet-list/example/index.scss
```diff
@@ -1,4 +1,4 @@
-@import "~@fremtind/jkl-core/src/typography/root.scss";
+@import "~@fremtind/jkl-core/core.scss";
```
lfbergee: import the entire core?

lfbergee: Ah, its example, thats fine

packages/core/package.json
```diff
@@ -18,10 +18,9 @@
     ],
     "scripts": {
         "build": "yarn build:styles && yarn build:modules",
-        "build:css": "node-sass src -o build/css --importer='../../node_modules/node-sass-tilde-importer'",
+        "build:styles": "node-sass . -o . --non-recursive --importer='../../node_modules/node-sass-tilde-importer'",
         "prebuild:scss": "mkdirp -p build && mkdirp build/scss",
```
lfbergee: This can go

packages/core/package.json
```diff
@@ -18,10 +18,9 @@
...
         "build:scss": "cpx 'src/**/*.scss' build/scss",
```
lfbergee: This too.

packages/radio-button/radio-button.scss
```diff
@@ -1,5 +1,4 @@
-@import "~@fremtind/jkl-core/build/scss/typography/font-size.scss";
-@import "~@fremtind/jkl-core/build/scss/variables/colors.scss";
+@import "~@fremtind/jkl-core/variables/variables";
```
lfbergee: Not a huge fan of the semantic `/variables/variables`.

Doesn't _index.scss work here? So you can do `@import "~@fremtind/jkl-core/variables";`

lfbergee:
Also cpx-depdendency can be deleted

packages/core/package.json
piofinn: Yep! That was me not remembering the underscore thing in Sass 😅

piofinn: Yep! And `--non-recursive` can be removed as well

packages/radio-button/radio-button.scss
piofinn: Maybe? I agree that it would be better, so I'll check it out!

piofinn: Seems like this was only implemented in LibSass in May, and is not a part of node-sass yet. Should be right around the corner, though!

https://github.com/sass/node-sass/issues/2339

packages/bullet-list/example/index.scss
piofinn: Yeah, I mostly just wanted to make sure it worked both ways!

piofinn:
> Also checkout the _index.scss file instead of _variables.scss and _mixin.scss

Not possible yet in node-sass as of yet (new feature in LibSass from May)

packages/radio-button/radio-button.scss
lfbergee: okay, how about all.scss?

import variables/all.scss?

Issue #124, opened 2019-07-03
Minified and uglified scripts


Saegrov:
As part of our build we should create a minified version as well.

PR #143, opened 2019-07-11
feat(build): add terser and size analyser to rollup

lfbergee:
## 📥 Proposed changes

Minifies all output from rollup. Maybe we want to keep commonjs non minified?

Add size-analyzer plugin to easier visualized bundle size of each packages.

## ☑️ Submission checklist

## 💬 Further comments
affects: @fremtind/jkl-accordion-react, @fremtind/jkl-bullet-list-react, @fremtind/jkl-button-react,
@fremtind/jkl-card-react, @fremtind/jkl-checkbox-react, @fremtind/jkl-core,
@fremtind/jkl-divider-line-react, @fremtind/jkl-dropdown-react, @fremtind/jkl-footer-react,
@fremtind/jkl-grid-react, @fremtind/jkl-header-react, @fremtind/jkl-logo-react,
@fremtind/jkl-radio-button-react, @fremtind/jkl-text-input-react, @fremtind/jkl-typography-react

ISSUES CLOSED: #124


lfbergee:
As of now this will break support for Safari 10 and IE8. IE8 is probably fine? Safari10 might not be?

Also we might want to set es5 as target?

piofinn:
> As of now this will break support for Safari 10 and IE8

Should be fine in Safari 10.1, so I would think is it fine? All computers able to run 10 are able to run 10.1, and caniuse put v10 share at 0,02%

lfbergee:
@piofinn I changed the rollup config a bit. Now it keeps the original built file, and also outputs a .min.js file. As of now noone will actually use that file, since the default path in package.json points to the non-minified file. We can change that if we want.

Issue #19, opened 2019-06-19
Toggle Switch prototype


Steinop:

Steinop:
Crap - Får ikke lastet opp MPG...

Steinop:
![Fremtind_Toggleswitch](https://user-images.githubusercontent.com/51952891/59767378-7c57d900-92a2-11e9-8aff-a7962f38eea6.gif)

Funker det med en GIF mon tro??

Steinop:
Grunnlag for knapper ligger i Figma på "Toggle Switch_Desktop size" Ser for meg at vi også bør lage en mobilverson tilpasset 17pt body tekst. https://www.figma.com/file/TkbB9ANfejDSjB2u4u1OEuqM/J%C3%B8kul-components?node-id=24%3A0

lfbergee:
I'll table this for now. We can use checkbox for this purpose, and we can get back to this when we have more components ready.

PR #178, opened 2019-07-22
Toggle switch 0

lfbergee:
## 📥 Proposed changes

Toggle switch style and react component

## ☑️ Submission checklist

## 💬 Further comments

This implementation is not as elegant as the checkbox, however this will work for older browsers as well.

CLOSES: #19


packages/toggle-switch/toggle-switch.scss
```diff
@@ -0,0 +1,66 @@
...
+$toggle-focus-outline-color: $himmel;
+
+.jkl-toggle-switch {
+    & input {
+        opacity: 0;
+        width: 0;
+        height: 0;
+    }
+    position: relative;
+    min-width: $toggle-width * 5;
```
piofinn: This should not be needed. I have a small refactor that gets rid of this and cleans up a few other things

packages/toggle-switch/toggle-switch.scss
```diff
@@ -0,0 +1,66 @@
...
+    & input:checked + .jkl-toggle-switch__slider:after {
+        transform: translateX(rem(30px)) scale(1.2);
+    }
+    html:not([data-mousenavigation]) & input:focus {
+        + .jkl-toggle-switch__slider {
+            box-shadow: 0 0 0 rem(2px) $snøhvit, 0 0 0 rem(4px) $toggle-focus-outline-color;
+        }
+    }
+    &__label {
+        position: absolute;
```
piofinn: No need to position this absolutely. Again, I have a fix for this in the refactor

lfbergee:
Yeah, I think i got a bit confused with different sketches, so yeah, i wasn't sure about the states. So did the most complicated thing variant, so we can simplify.

packages/toggle-switch/toggle-switch.scss
lfbergee: Cool, yeah, this should be aviodable

packages/toggle-switch-react/src/ToggleSwitch.tsx
```diff
@@ -0,0 +1,17 @@
+import React, { ReactNode } from "react";
+
+interface Props {
```
piofinn: Shouldn't there be an onChange prop?

lfbergee: Probably, I work out from checkbox, and it does not have it. But yeah, it should be easy to hook into from a parent component

piofinn: Huh, probably just a slip-up in checkbox as well. In my mind both will be used as controlled inputs more often than not

packages/toggle-switch-react/src/ToggleSwitch.tsx
```diff
@@ -4,12 +4,18 @@ interface Props {
...
-export const ToggleSwitch = ({ children, defaultChecked, className = "" }: Props) => {
+export const ToggleSwitch = ({ children, defaultChecked, onChange, className = "" }: Props) => {
     return (
         <label data-testid="jkl-toggle-input--label" className={`jkl-toggle-switch ${className}`}>
-            <input data-testid="jkl-toggle-input" type="checkbox" defaultChecked={defaultChecked} />
+            <input
+                data-testid="jkl-toggle-input"
+                type="checkbox"
+                defaultChecked={defaultChecked}
+                onChange={(e) => onChange && onChange(e.target.checked)}
```
lfbergee: Should be a named function

packages/toggle-switch-react/src/ToggleSwitch.test.tsx
```diff
@@ -0,0 +1,50 @@
+import React from "react";
```
lfbergee: We should add a test for the onchange handler

piofinn: Done :)

Issue #112, opened 2019-07-03
Font does not work in Safari


piofinn:
Safari doen't let you access locally installed fonts (other than system fonts) to prevent fingerprinting. We therefore have to host the font files if we want to be able to preview our work in Safari.

## Proposed solution

- [ ] Find somewhere to host the font files
- [ ] Change the `typography/core.scss` file to use these files as `src`

Issue #37, opened 2019-06-21
Bundle font


lfbergee:
- [ ] Font should be bundled with jøkul
- [ ] Add font lisence and usage documentation

PR #179, opened 2019-07-22
fix(core): ship font inline

lfbergee:
affects: @fremtind/jkl-core

## 📥 Proposed changes

base64 encode font into css file, should be hosted when the font font arrives

## ☑️ Submission checklist

## 💬 Further comments

CLOSES: #112 
CLOSES: #37


lfbergee:
Added the regular font as well, and used font-weight values from before.

PR #181, opened 2019-07-22
Css fixes 2

lfbergee:
## 📥 Proposed changes

- Buttons had the rather ugly dottet line around on firefox.
- Wierd different color background on select in datepicker

## ☑️ Submission checklist


packages/datepicker/datepicker.scss
```diff
@@ -47,10 +47,12 @@ $background-color: $snøhvit;
                 display: block;
                 box-shadow: inset 0 rem(-2px) 0 0 $svart;
                 transition: box-shadow 175ms ease-in;
+                background-color: $snøhvit;
+                text-transform: capitalize;
 
                 font-size: 1.75rem;
                 line-height: 3rem;
-                padding-bottom: $component-spacing--small;
+                height: rem(50px);
```
piofinn: Why 50px? If we're consolidating this with text-input and dropdown it should be 40px (32px line height + 8px bottom padding)

lfbergee: Actually the height of the input field is 48px, so it should be 48, not 50

lfbergee: But that does not look right, 50px does look correct.

piofinn: Huh, OK.. I was looking at the dropdown style, where it is 40px (plus bottom border). We definitely need to have the designers take a look at these components to find common values 😅

Issue #117, opened 2019-07-03
Components: Buttons - Desktop size


Steinop:
Bigger button for desktop. The original button has 17 px text size. Increase to 22px body size for desktop and 56 px (?) button height. Increased padding must be harmonized with the original button.

Need for Primary, Secondary and Tertiary buttons.

PR #190, opened 2019-07-23
fix(button): add larger desktop buttons

lfbergee:
affects: @fremtind/jkl-button-react, @fremtind/jkl-button

ISSUES CLOSED: #117

## 📥 Proposed changes

Larger desktop buttons

## ☑️ Submission checklist


packages/button/button.scss
```diff
@@ -9,21 +11,32 @@ $button-focus-ring-color: $himmel;
...
-$button-height: 2.5rem;
-$button-width: $button-height;
 $button-border-width: rem(1px);
 $animation-timing: 100ms cubic-bezier(0.6, 0.2, 0.35, 1);
 $hover-elevation-distance: -0.25rem;
 
+$button-height--large: 3.5rem;
+$button-height--small: 2.5rem;
+
+@mixin deviceScale() {
```
piofinn: Does this need to be a mixin? Feels like it would be clearer if it was just  part of the class style..

piofinn:
These are some massive buttons! 😅 I think maybe 48px looks better (and still quite large), but we should hear with the designers when they get back

lfbergee:
Yes, its big. I'll merge, and we can revisit this when ux is back

PR #191, opened 2019-07-23
fix(dropdown): fixed minor placement issue, update documentation

lfbergee:
affects: @fremtind/jkl-dropdown-react, @fremtind/jkl-dropdown

## 📥 Proposed changes

Dropdown was place a little to low when no value were selected. Imports did not work on linux. Updated readmes to updated standard.

## ☑️ Submission checklist

### DIFF
--- a/packages/dropdown/dropdown.scss
+++ b/packages/dropdown/dropdown.scss
@@ -18,7 +18,7 @@ $side-padding: $component-spacing--small;
 $bottom-padding: $component-spacing--small;
 $bottom-margin: $component-spacing--large;
 
-$dropdown-total-height: rem(68px);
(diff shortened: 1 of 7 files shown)


packages/dropdown/dropdown.scss
```diff
@@ -18,6 +18,8 @@ $side-padding: $component-spacing--small;
 $bottom-padding: $component-spacing--small;
 $bottom-margin: $component-spacing--large;
 
+$dropdown-total-height: rem(68px);
```
piofinn: How about `$label-height + $value-height + $bottom-padding`? Makes it super clear and updates automagically 😎

lfbergee: Then it will cover the bottom line, but that kinda looks better. So I've updated it.

PR #204, opened 2019-07-26
fix(text-input): show underline in mobile Safari

piofinn:
## 📥 Proposed changes

Specify `-webkit-appearance: none` on text input fields to ensure bottom border is shown in mobile Safari 🤷‍♂️

## ☑️ Submission checklist


piofinn:
We could probably use just `appearance: none` and let autoprefixer deal with it, but I'm also a little afraid that it will introduce new bugs. Yay, CSS! 🎉

PR #217, opened 2019-07-26
Add style and component for links

piofinn:
## 📥 Proposed changes

Add style and component for links

## ☑️ Submission checklist


packages/core/links.scss
```diff
@@ -0,0 +1,33 @@
+@import "./variables/all";
+@import "./functions";
+
+.jkl-link {
+    appearance: none;
+    outline: none;
+    :-moz-focusring {
```
lfbergee: Could we place this rather annoying little bastard in normalize?

packages/typography-react/src/Link.tsx
```diff
@@ -0,0 +1,16 @@
+import React, { ReactNode } from "react";
+
+interface Props {
+    href: string;
+    inline?: boolean;
+    children: ReactNode;
+    className?: string;
+    target?: string;
```
lfbergee: Maybe we should limit this to valid arguments?

lfbergee: "_self" | "_blank" | "_parent" | "_top"

Issue #41, opened 2019-06-21
Components: Field status message


lfbergee:
- [x] Initial design
- [x] Initial package

Component to go under  form-elements to inform about error, missing information and such.

Steinop:
https://www.figma.com/file/TkbB9ANfejDSjB2u4u1OEuqM/J%C3%B8kul-components?node-id=111%3A9

lfbergee:
I'm not sure this should be it's own component anymore. The field message basicly do not have any particular design elements, it's just a paragraph, so I'm not sure its worth the extra complexity to have this a it's own components, instead it can be a part of whatever components that needs it.

Oh, and the error color do not look WCAG compliant to my eyes.

lfbergee:
I think we should add the style to core, that way any component that needs it can pick the style from there.

Somethink like:
```
.jkl-form__help-text{}
.jkl-form__error-text{}
```

PR #234, opened 2019-08-01
Help text

lfbergee:
## 📥 Proposed changes

Add help and error text to input field and area. Add error color to core. Fixed minor ios issues in portal and hamburger.

## ☑️ Submission checklist

## 💬 Further comments

CLOSES: #41


packages/core/help-text.scss
```diff
@@ -0,0 +1,17 @@
+@import "./variables/all";
+@import "./mixins/all";
+
+.jkl {
+    &__text {
```
lfbergee: Better nameing suggestion?

piofinn: Alternatively, we could do something a bit more abstract that would let us use component-specific BEM naming in the different form fields:
```scss
[class$="__message"] {
    // common styles
    &--help {
        color: $svart;
    }
    &--error {
        color: $error-color;
    }
}
```
Then in `text-field` we could do
```jsx
<span className="jkl-text-field__message jkl-text-field__message--error">Feilmelding</span>
```
and similarly in  `dropdown`
```jsx
<span className="jkl-dropdown__message jkl-dropdown__message--error">Feilmelding</span>
```
and so on. The discoverability for the developers might be a bit worse, though..

packages/text-input/text-input.scss
```diff
@@ -18,18 +18,15 @@ $bottom-margin: $component-spacing--large;
...
     display: block;
     box-sizing: border-box;
     position: relative;
     font-family: inherit;
     width: 100%;
-    height: $input-total-height;
+    min-height: $input-total-height;
     padding-top: $label-height;
     background-color: $input-bg-color;
-    margin-bottom: $component-spacing--large;
```
piofinn: I don't think we should get rid of the margin. Having the spacing as a margin (like now) acts more like a "minimum spacing", as it won't add up with adjacent margins. In the unlikely event that there needs to be less than `1rem` spacing between the input and the element below, it can be overridden by the new spacing API classes.

packages/text-input-react/src/TextArea.tsx
```diff
@@ -8,11 +8,15 @@ interface Props {
...
+    helpText?: string;
+    errorText?: string;
 }
 
-export const TextArea = ({ isInvalid = false, id, label, className = "", ...rest }: Props) => (
+export const TextArea = ({ isInvalid = false, id, label, className = "", helpText, errorText, ...rest }: Props) => (
     <label data-testid="jkl-text-field" className={`jkl-text-field jkl-text-area ${className}`}>
         <textarea aria-invalid={isInvalid} className="jkl-text-field__input" id={id} placeholder=" " {...rest} />
         <span className="jkl-text-field__label">{label}</span>
+        {helpText && <span className="jkl__text jkl__text--help">{helpText}</span>}
```
piofinn: Help text should be replaced by error text, so we need `&& !(isInvalid && errorText)` here

lfbergee: i fix

packages/core/help-text.scss
lfbergee: I do like the last one, but it seems a bit magical i think.

How about `jkl-form-support-text`?

packages/core/core.scss
```diff
@@ -4,4 +4,5 @@
 @import "./headings.scss";
 @import "./paragraphs.scss";
 @import "./links.scss";
+@import "./help-text.scss";
```
piofinn: I'm thinking maybe we should put this, paragraph, and headings into one file? Maybe `typography.scss`?

If not, it should probably be called `form-support-text.scss` since we renamed the class 😊

packages/text-input-react/src/SupportText.tsx
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+
+interface Props {
+    helpText?: string;
+    errorText?: string;
+    isInvalid: boolean;
+}
+
+export const SupportText = ({ helpText, errorText, isInvalid }: Props) => {
```
piofinn: If we've made this into a component, should it be its own package in the library?

lfbergee: Yeah, but I think that should be refactored when another component wants to use it

packages/core/core.scss
lfbergee: Good idea, will fix

PR #284, opened 2019-08-06
Animate accordion

Saegrov:
## 📥 Proposed changes

Add animation to accordion

## ☑️ Submission checklist


Saegrov:
> Nice, great work!
> 
> Could have been nice with a unit test, but then again, I saw the talk yesterday and saw there was a difference when invoking a microtask from js instead of an button click, so it might be a bit awkard to test.

I tried but since we use jsdom for our tests it is hard/impossible to check if the height changes (since jsdom doesn't do layout and height calculation).

PR #324, opened 2019-08-15
Datepicker fixes

Saegrov:
## 📥 Proposed changes

* Nudge the datepicker 2px down to conform with the design.
* Close datepicker when choosing date
* Add ability to type text into datepicker 
* Add bottomtext to datepicker
* Add label to text-field (use by datepicker)
* Change opening behaviour to opening datepicker on focus
* Capitalize months in datepicker

## ☑️ Submission checklist


Saegrov:
This change does have a bug in chrome: https://github.com/nrkno/core-components/issues/322

piofinn:
I've found one bug/unexpected behavior (that might have been there before as well): The dropdown stays open when you tab to another input, possibly blocking that input or others. This is especially relevant when entering a date as text and then tabbing to the next input in the form.

packages/datepicker-react/src/DatePicker.tsx
```diff
@@ -74,7 +99,63 @@ export function DatePicker({
...
+        if (ref.current && !ref.current.contains(e.target as Node)) {
+            onClickOutside();
+        }
+    }
+
+    useEffect(() => {
+        // Bind the event listener
+        document.addEventListener("mousedown", onMouseDown);
+        // @ts-ignore
+        document.addEventListener("focusin", onMouseDown);
```
piofinn: `@ts-ignore` is unfortunately required, as the `focusin` event is not yet registered in the `FocusEvent` interface. The event type is supported by all browsers we should consider, and should be patched into TypeScript soon (see https://github.com/Microsoft/TypeScript/issues/30716).

Saegrov: @piofinn 
I think we should add a reference to that issue in the // @ts-ignore. Otherwise - that comment will never be fixed.

packages/text-input/text-input.scss
```diff
@@ -27,13 +27,31 @@ $textarea-height--expanded: $input-height * 7 + $bottom-padding;
...
+    &__placeholder {
+        @include font-size(small);
+        @include line-height(small);
+
+        color: transparent;
+        user-select: none;
+        left: 0;
+        padding-left: 0.125rem;
+        position: absolute;
+        transform: translate3d(0, 20px, 0);
```
piofinn: Why 3d? And should the px value be a variable?

packages/datepicker-react/src/DatePicker.tsx
```diff
@@ -1,10 +1,9 @@
...
+import { Select } from "@fremtind/jkl-dropdown-react";
+import { TextField } from "@fremtind/jkl-text-input-react";
+import { SupportLabel } from "@fremtind/jkl-typography-react";
 // @ts-ignore
 import CoreDatepicker from "@nrk/core-datepicker/jsx";
-// @ts-ignore
-import CoreToggle from "@nrk/core-toggle/jsx";
-import { TextField } from "@fremtind/jkl-text-input-react";
-import { Select } from "@fremtind/jkl-dropdown-react";
+import React, { ChangeEvent, MutableRefObject, useEffect, useRef, useState } from "react";
```
piofinn: Can we move this to the top? My OCD is kicking in when external deps come after internal ones 😅

packages/datepicker-react/src/DatePicker.tsx
```diff
@@ -18,53 +17,79 @@ interface Props {
...
 }: Props) {
-    const [today] = useState(Date.now() - (Date.now() % 864e3));
     const [date, setDate] = useState(initialDate);
-    const [showPicker, togglePicker] = useState(initialShow);
+    const [datepickerHidden, setDatepickerHidden] = useState(!initialShow);
+    const [dateString, setDateString] = useState(initialDate ? formatDate(initialDate) : "");
+    const ref = useOnClickOutside<HTMLDivElement>(() => setDatepickerHidden(true));
+
+    function onInputChange(event: ChangeEvent<HTMLInputElement>) {
+        const newString = event.target.value;
```
piofinn: `newDateString`?

packages/datepicker-react/src/DatePicker.tsx
```diff
@@ -108,7 +108,7 @@ export function DatePicker({
 function useOnClickOutside<T extends HTMLElement>(onClickOutside: () => void): MutableRefObject<T | null> {
```
Saegrov: This should be renamed since the functionality has been changed.

packages/datepicker-react/src/DatePicker.tsx
```diff
@@ -108,7 +108,7 @@ export function DatePicker({
...
     const ref = useRef<T>(null);
 
-    function onMouseDown(e: MouseEvent) {
+    function onMouseDown(e: MouseEvent | FocusEvent) {
```
Saegrov: This function should also be renamed.

packages/datepicker-react/src/DatePicker.tsx
```diff
@@ -21,6 +22,8 @@ interface Props {
     onlyFuture?: boolean;
     initialShow?: boolean;
     className?: string;
+    bottomText?: string;
```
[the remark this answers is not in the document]
Saegrov: Agree that this is not a great name. SupportText is also a bit confusing imo. How about bottomLabel?

packages/datepicker-react/src/DatePicker.tsx
Saegrov: Move away :)

packages/datepicker-react/src/DatePicker.tsx
Saegrov: Yes!

packages/datepicker-react/src/DatePicker.tsx
```diff
@@ -18,53 +17,79 @@ interface Props {
...
-            </button>
-            <CoreToggle hidden={!showPicker} popup onToggle={toggle}>
+        <div className={`jkl-datepicker ${className}`} ref={ref}>
+            <TextField
+                placeholder={"dd.mm.åååå"}
+                label={label}
+                type="text"
+                value={dateString}
+                onChange={onInputChange}
+                onFocus={() => setDatepickerHidden(false)}
```
lfbergee: Lets pick this out to avoid unnessesary rerenders

packages/datepicker-react/src/DatePicker.tsx
```diff
@@ -18,53 +17,79 @@ interface Props {
...
-                    type="text"
-                    readOnly
-                    value={date.toLocaleDateString()}
-                    data-testid="jkl-datepicker-input"
-                />
-            </button>
-            <CoreToggle hidden={!showPicker} popup onToggle={toggle}>
+        <div className={`jkl-datepicker ${className}`} ref={ref}>
+            <TextField
+                placeholder={"dd.mm.åååå"}
```
lfbergee: Should probably be a prop with this as default value?

packages/datepicker-react/src/DatePicker.tsx
```diff
@@ -74,7 +99,63 @@ export function DatePicker({
...
+
+    function onMouseDown(e: MouseEvent | FocusEvent) {
+        if (ref.current && !ref.current.contains(e.target as Node)) {
+            onClickOutside();
+        }
+    }
+
+    useEffect(() => {
+        // Bind the event listener
+        document.addEventListener("mousedown", onMouseDown);
```
lfbergee: Should check document before adding eventlisnter

packages/datepicker-react/src/DatePicker.tsx
lfbergee: Can we use the clickoutside hoc we have? And create another one for focus-loss perhaps?

packages/datepicker-react/src/DatePicker.tsx
```diff
@@ -74,7 +99,63 @@ export function DatePicker({
...
+            // Unbind the event listener on clean up
+            document.removeEventListener("mousedown", onMouseDown);
+            // @ts-ignore
+            document.removeEventListener("focusin", onMouseDown);
+        };
+    }, []);
+    return ref;
+}
+
+function isSameDay(date1: Date, date2: Date) {
```
lfbergee: maybe pendantic, but i'd like to see a test or two for each of these.

packages/datepicker-react/src/DatePicker.tsx
```diff
@@ -74,7 +99,63 @@ export function DatePicker({
...
+    );
+}
+
+/**
+ *
+ * @param date the date to format
+ * @return returns a date with "dd.mm.yyyy"-format
+ */
+function formatDate(date: Date) {
+    const day = pad0(date.getDate());
```
lfbergee: `date.getDate().padStart(2,"0")` ES7 FTW!

packages/text-input/text-input.scss
```diff
@@ -27,13 +27,31 @@ $textarea-height--expanded: $input-height * 7 + $bottom-padding;
...
     background-color: $input-bg-color;
-    padding-bottom: $component-spacing--large;
 
     &--inline {
         display: inline-block;
         vertical-align: bottom;
     }
 
+    &__input:focus ~ .jkl-text-field__placeholder {
+        color: darkgray;
```
lfbergee: Shouldn't this be one of our colors? And if its missing, maybe a FIXME or something to update it.

packages/datepicker-react/src/DatePicker.tsx
piofinn: Not quite the same, though. Unlike the current function it will pad 0 to 00, but I guess that should never be tha case anyway 🤷‍♂️

piofinn: Also, I don't think padStart accepts numbers?

packages/datepicker-react/src/DatePicker.tsx
```diff
@@ -14,57 +14,91 @@ interface Props {
...
-    const [today] = useState(Date.now() - (Date.now() % 864e3));
     const [date, setDate] = useState(initialDate);
-    const [showPicker, togglePicker] = useState(initialShow);
+    const [datepickerHidden, setDatepickerHidden] = useState(!initialShow);
+    const [dateString, setDateString] = useState(initialDate ? formatDate(initialDate) : "");
+    const openDatepicker = () => setDatepickerHidden(false);
+    const closeDatepicker = () => setDatepickerHidden(true);
+
+    const datepickerRef = useRef<HTMLDivElement>(null);
+    useClickOutside(datepickerRef, closeDatepicker);
```
lfbergee: Cool! I like this a lot more. I think we should consider creating a compound hook with all of these if we see more components that has the same needs. 

Functional and beautiful!

packages/datepicker-react/src/DatePicker.tsx
lfbergee: `0` could never happen, if so, shit has hit the fan. But yeah, number is not valid, ``${date.getDate()}`.pad(2,"0")` should be fine.

You decide, but I would prefer a unit test for pad0(if we keep it) and formatDate.

packages/datepicker-react/src/DatePicker.tsx
```diff
@@ -74,7 +108,38 @@ export function DatePicker({
...
+export function formatDate(date: Date) {
+    const day = pad0(date.getDate());
+    const month = pad0(date.getMonth() + 1);
+    return `${day}.${month}.${date.getFullYear()}`;
+}
+
+/**
+ * @return returns a string with 0 in front of single digit numbers
+ */
+function pad0(nr: number) {
```
lfbergee: Also, I might be in the minority here. Its pretty to define the function after usage, but I strongly dislike it. To me it feels like abuse of language design flaw. I.E someone wants to refactor into an arrow functions, and it just stops working.

packages/react-hooks/src/useFocusOutside.ts
```diff
@@ -0,0 +1,24 @@
...
+        if (ref && ref.current && !ref.current.contains(event.target as Node)) {
+            fn();
+        }
+    }
+
+    React.useEffect(() => {
+        if (ref) {
+            // focusin event not yet typed. See https://github.com/microsoft/TypeScript/issues/30716
+            // @ts-ignore
+            document.addEventListener("focusin", handleFocusOutside);
```
lfbergee: Document should probably always be fine here, but I would feel a it more comfortable with a check.

packages/text-input-react/src/TextField.tsx
```diff
@@ -39,8 +43,10 @@ export const TextField = ({
             id={id}
             placeholder=" "
             readOnly={readOnly}
+            value={value}
             {...rest}
         />
+        {!value && <span className="jkl-text-field__placeholder">{placeholder}</span>}
```
lfbergee: maybe a default value for placeholder and/or a check if placeholder exists before usage?

packages/text-input/text-input.scss
lfbergee: ☝️

Issue #330, opened 2019-08-19
Labels for Checkboxes and Radiobuttons vs labels for Dropdowns and Input fields


Steinop:
Case for "SeOpp - Melde Skade"

In some cases a dropdown label can be written as a question which will give the minimised label text a reduced readability due to a smaller font size, gray colour and longer lines of text.

Steinop:
Figma reference for Input label v2.0:

Normal size: 
https://www.figma.com/file/Wh8Gfnesn25shH31sKyIaT/J%C3%B8kul-bibliotek?node-id=559%3A1519

Small size:
https://www.figma.com/file/Wh8Gfnesn25shH31sKyIaT/J%C3%B8kul-bibliotek?node-id=563%3A1626

Steinop:
Animation Preview:
![Fremtind_Inputfelt_Aug2019](https://user-images.githubusercontent.com/51952891/64165623-bc603e00-ce45-11e9-8c14-bd5a27bf59f7.gif)

Mikkelinski:
Solved in separate issue

PR #336, opened 2019-08-26
Checkbox group

### LINKED ISSUE #314: Checkbox group
Basically the same as radiobutton-group but for checkbox.

piofinn:
## 📥 Proposed changes

~~Add checkbox group component and style based on RadioButtonChoice.~~

- Add a FieldGroup component with related style, to use when creating groups of inputs in forms.
- Add a CheckboxGroup component that utilizes FieldGroup
- Update RadioButtonGroup to utilize FieldGroup

## ☑️ Submission checklist

## 💬 Further comments

Closes #314


packages/checkbox-react/src/CheckboxGroup.tsx
```diff
@@ -0,0 +1,21 @@
+import React from "react";
+import { Checkbox } from "./Checkbox";
+
+interface Props {
+    legend: string;
+    choices: string[];
```
piofinn: We might want to be able to pass both name and label value for each option/choice. Maybe something like
```tsx
interface Choice {
    name: string;
    label: string;
}
interface Props {
    /* ... */
    choices: Choice[] | string[];
}
```

packages/checkbox-react/example/index.tsx
```diff
@@ -3,25 +3,26 @@ import { initTabListener } from "@fremtind/jkl-core";
...
 
 const App = () => {
     const [isCool, changeCoolness] = React.useState(false);
     return (
         <>
-            <Checkbox>I am checkbox!</Checkbox>
-            <Checkbox>Am I?</Checkbox>
-            <Checkbox checked={true}>I will never change</Checkbox>
-            <Checkbox checked={isCool} onChange={() => changeCoolness(!isCool)}>
+            <Checkbox name="checkbox1" onChange={(name, value) => alert(`${name} is now ${value}`)}>
```
Saegrov: I really don't understand why we use `alert`? I think it is super annoying! The same applies for the portal.

packages/checkbox/checkbox.scss
```diff
@@ -73,6 +77,20 @@ $checkbox-focus-outline-color: $himmel;
     }
 }
 
+.jkl-checkbox-group {
+    border: 0;
+    padding: 0;
+    margin: 0;
+    line-height: 1;
+
+    & > legend {
```
Saegrov: Use BEM.

portal/src/examples/CheckboxExample.tsx
```diff
@@ -1,27 +1,34 @@
...
-        <Checkbox onChange={(val) => alert(val)}>Do you like mulitple choices?</Checkbox>
-    </div>
+    <Checkbox name="I am uncontrolled" />
+    <Checkbox name="controlled" checked={isChecked} onChange={()=>setChecked(!isChecked)}>I am controlled</Checkbox>
+    <Checkbox name="locked" checked={true}>I always checked</Checkbox>
+    <CheckboxGroup
+        className="jkl-spacing--top-4"
+        inline={false}
+        choices={["I am checkbox", "Do you like multiple choices?"]}
+        onChange={(name, val) => alert(name + " is now " + val)}
```
Saegrov: Why do we use `alert`? It is so annoying.

packages/checkbox/checkbox.scss
```diff
@@ -73,6 +77,20 @@ $checkbox-focus-outline-color: $himmel;
...
+        font-size: $font-size-2;
+        line-height: $line-height-2;
```
Saegrov: Should we use one of the mixins for this? Does it have the same style as some other text?

gulpfile.js
```diff
@@ -10,12 +10,16 @@ sass.compiler = require("node-sass");
 /* eslint-enable @typescript-eslint/no-var-requires */
 
 const scssFiles = ["**/*.scss", "!example/*.scss"];
+function throwError(e) {
```
Saegrov: throwSassError?

packages/checkbox/checkbox.scss
piofinn: These will probably have to be changed anyways (see #330)

packages/checkbox-react/src/CheckboxGroup.tsx
```diff
@@ -0,0 +1,16 @@
+import React, { ReactNode } from "react";
+import { FieldGroup } from "@fremtind/jkl-field-group-react";
+
+interface Props {
+    children: ReactNode;
+    legend: string;
+    className?: string;
+}
+
+export function CheckboxGroup({ legend, className, children }: Props) {
```
Saegrov: Imo CheckboxGroup should have the same public API as RadioButtonGroup.
Instead of taking children (which must be Checkboxes) as props, it should take an array of strings which should map to Checkboxes.

packages/field-group-react/src/FieldGroup.tsx
```diff
@@ -0,0 +1,14 @@
+import React, { ReactNode } from "react";
+
+interface Props {
+    legend: string;
+    children?: ReactNode;
+    className?: string;
+}
+
+export const FieldGroup = ({ legend, className, children }: Props) => (
```
Saegrov: Do we really need this FieldGroup? I don't feel it creates enough value to a be a separate react-component? It feels too DRY.

Saegrov: If we look at CheckboxGroup now it doesn't do anything. It's just a Component that wraps a Component. Or a function that calls another function.

packages/checkbox-react/src/Checkbox.tsx
```diff
@@ -1,25 +1,28 @@
...
                 className="jkl-checkbox__input"
                 data-testid="jkl-checkbox-input"
                 checked={checked}
                 type="checkbox"
+                name={name}
                 onChange={handleChange}
             />
             <span className="jkl-checkbox__check-mark" />
-            {children}
+            {children ? children : name}
```
Saegrov: This feels like a weird api:
```
<Checkbox name="Will this be displayed?" />
            <Checkbox name="Will this be displayed">Or this be displayed?</Checkbox>
```

piofinn: Yeah, that was mostly there to accommodate a string array of options in CheckboxGroup in addition to a more HTML-like implementation. Should probably be reverted.

packages/field-group-react/src/FieldGroup.tsx
piofinn: I would flip it around and ask, do we really need the CheckboxGroup? We have a couple of alternatives:
- We can have the same API in CheckboxGroup and RadioButtonChoice, and implement the same wrapper and style twice
- We can have different APIs, but otherwise same as above
- We can have a separate FieldGroup with style and structure and use it standalone and in other components

I don't think the first option is right, since the resulting code is not DRY and these are two quite different concepts: With radio buttons you choose _between_ options, and as such the choices can be considered as one thing (and in fact _are_ considered one thing in standard HTML forms). With checkboxes however, the boxes each represent one choice, and the group is there to visually and semantically connect them.

The second option.. 🤷‍♂️ 

The third option is how it is implemented here. I agree that the CheckboxGroup might be a too thin wrapper around FieldGroup to be considered useful, so maybe it is not needed?

gulpfile.js
```diff
@@ -10,12 +10,16 @@ sass.compiler = require("node-sass");
 /* eslint-enable @typescript-eslint/no-var-requires */
 
 const scssFiles = ["**/*.scss", "!example/*.scss"];
+function throwSassError(e) {
+    sass.logError(e);
+    throw new Error("sass compilation failed", e);
```
lfbergee: We could probably drop the e here, as it is logged much more readable in the line above. For the furture, we throw an error here to fail the build process if compilation fails. sass.logError only logs, it does not stop the build.

packages/checkbox-react/example/index.tsx
```diff
@@ -1,4 +1,5 @@
-import "@fremtind/jkl-checkbox/checkbox.css";
+import "@fremtind/jkl-checkbox/checkbox.min.css";
+import "@fremtind/jkl-field-group/field-group.min.css";
 import { initTabListener } from "@fremtind/jkl-core";
 import "@fremtind/jkl-core/core.css";
 import React from "react";
```
lfbergee: Does this not annoy your ocd? I can feel it.

packages/checkbox-react/src/Checkbox.tsx
```diff
@@ -1,25 +1,28 @@
...
     className?: string;
 }
 
-export function Checkbox({ children, checked, onChange, className = "" }: Props) {
-    const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => onChange && onChange(e.target.checked);
+export function Checkbox({ children, name, checked, onChange, className = "", inline = true }: Props) {
+    const handleChange = (e: ChangeEvent<HTMLInputElement>) => onChange && onChange(name, e.target.checked);
     return (
-        <label className={`jkl-checkbox ${className}`}>
+        <label className={`jkl-checkbox ${inline ? "jkl-checkbox--inline" : ""} ${className}`}>
```
lfbergee: I kinda liked the version from another component

```
"jkl".concat(inline ? " jkl--inline" : "");
```

packages/checkbox-react/src/Checkbox.tsx
lfbergee: Should this still be here? I'm not a huge fan of this either.

packages/field-group-react/README.md
```diff
@@ -0,0 +1,9 @@
+# [`@fremtind/field-group`](https://fremtind.github.io/jokul/components/fieldgroup/)
```
lfbergee: This is not a component in the portal, so this link will be broken

packages/field-group-react/README.md
```diff
@@ -0,0 +1,9 @@
...
+
+## Beskrivelse
+
+[Bruk og prinsipper](https://fremtind.github.io/jokul/components/dropdown/) er beskrevet i `@fremtind/jkl-dropdown` stil-pakken.
```
lfbergee: dropdown coppypasta

packages/checkbox-react/src/Checkbox.tsx
piofinn: Oops, missed that one!

packages/field-group-react/src/FieldGroup.test.tsx
```diff
@@ -0,0 +1,31 @@
+import React from "react";
```
lfbergee: 👍

packages/checkbox-react/example/index.tsx
piofinn: Haha!

PR #337, opened 2019-08-26
feat(autosuggest): autosuggest / searchable dropdown component

Martinvks:
affects: @fremtind/jkl-autosuggest-react, @fremtind/jkl-autosuggest

## 📥 Proposed changes

Introduce a searchable dropdown / autosuggest / autocomplete component.

## ☑️ Submission checklist

## 💬 Further comments

If accepted this pr will add two new external dependencies:
https://github.com/downshift-js/downshift
https://github.com/kentcdodds/match-sorter

I did not have any design when creating this component, so any feedback on improvements is welcome.


lfbergee:
Cool, thanks for input. However, I do think this is a bit redundant, as the Dropdown component could support typeahead and seach with some adjustments. I would prefer this to be contained in that component, or for it to be and component that extends the normal dropdown. That way we would avoid introducing a new dependency, with a new API and a new set of style to be maintained.

I'd at least like us to explore that path before merging this.

Looks like a lot of great work, hope you found Jøkul nice to work with, we'd love some feedback on the developer experience.

lfbergee:
@all-contributors please add @Martinvks for code

PR #363, opened 2019-09-12
New table component

Saegrov:
## 📥 Proposed changes

This introduces a new table component.
As a first version it is meant only as a simple data table, with support to have a row act as a link.

## ☑️ Submission checklist

### DIFF
--- a/packages/table-react/src/Table.tsx
+++ b/packages/table-react/src/Table.tsx
@@ -1,0 +1,30 @@
+import React from "react";
+import { TableAnchorRowData, TableRow } from "./TableRow";
+
+export type TableRowData = string[] | TableAnchorRowData;
+
+interface Props {
+    columns: string[];
+    rows: Array<TableRowData>;
+}
+
+export function Table(props: Props) {
+    return (
+        <table className="jkl-table">
+            <thead>
+                <tr className="jkl-table__row">
+                    {props.columns.map((columnValue, i) => (
+                        <td className="jkl-table__heading" key={i}>
+                            {columnValue}
+                        </td>
+                    ))}
+                </tr>
+            </thead>
+            <tbody>
+                {props.rows.map((row, i) => (
+                    <TableRow row={row} key={i} />
+                ))}
+            </tbody>
+        </table>
+    );
+}
--- a/packages/table-react/src/Table.tsx
+++ b/packages/table-react/src/Table.tsx
@@ -14,9 +14,9 @@ export function Table(props: Props) {
             <thead>
                 <tr className="jkl-table__row">
                     {props.columns.map((columnValue, i) => (
-                        <td className="jkl-table__heading" key={i}>
+                        <th className="jkl-table__heading" key={i}>
                             {columnValue}
-                        </td>
+                        </th>
                     ))}
                 </tr>
             </thead>
--- a/packages/table-react/src/TableRow.tsx
+++ b/packages/table-react/src/TableRow.tsx
@@ -13,7 +13,7 @@ export interface TableAnchorRowData extends BaseTableRowData {
     type: "anchor";
     href: string;
     hrefLabel: string;
-    onClickRow?: (href: string) => void;
+    onRowClick?: (href: string) => void;
 }
 
 interface Props {
@@ -29,7 +29,7 @@ export function TableRow(props: Props) {
 
     if (isAnchorRowData(row)) {
         onClick = () => {
-            row.onClickRow && row.onClickRow(row.href);
+            row.onRowClick && row.onRowClick(row.href);
         };
         rowModifierClasses += "jkl-table__row--anchor-row";
     }
--- a/packages/table-react/src/TableRow.tsx
+++ b/packages/table-react/src/TableRow.tsx
@@ -29,7 +29,11 @@ export function TableRow(props: Props) {
 
     if (isAnchorRowData(row)) {
         onClick = () => {
-            row.onRowClick && row.onRowClick(row.href);
+            if (row.onRowClick) {
+                row.onRowClick(row.href);
+            } else {
+                window.location.href = row.href;
+            }
         };
         rowModifierClasses += "jkl-table__row--anchor-row";
     }
--- a/packages/table-react/src/TableRow.tsx
+++ b/packages/table-react/src/TableRow.tsx
@@ -1,0 +1,48 @@
+import React, { MouseEvent } from "react";
+import { TableRowData } from "./Table";
+import { TableData } from "./TableData";
+
+export type TableRowDataVariants = "anchor";
+
+export interface BaseTableRowData {
+    type: TableRowDataVariants;
+}
+
+export interface TableAnchorRowData extends BaseTableRowData {
+    rowData: string[];
+    type: "anchor";
+    href: string;
+    hrefLabel: string;
+    onClickRow?: (href: string) => void;
+}
+
+interface Props {
+    row: TableRowData;
+}
+
+export function TableRow(props: Props) {
+    const { row } = props;
+    const rowData: string[] = isAnchorRowData(row) ? row.rowData : row;
+
+    let onClick: undefined | ((evt: MouseEvent<HTMLTableRowElement>) => void) = undefined;
+    let rowModifierClasses = "";
+
+    if (isAnchorRowData(row)) {
+        onClick = () => {
+            row.onClickRow && row.onClickRow(row.href);
+        };
+        rowModifierClasses += "jkl-table__row--anchor-row";
+    }
+
+    return (
+        <tr onClick={onClick} className={`jkl-table__row jkl-table__row--data-row ${rowModifierClasses}`}>
+            {rowData.map((data, j) => (
+                <TableData data={data} row={row} key={j} isFirstCell={j === 0} />
+            ))}
+        </tr>
+    );
+}
+
+export function isAnchorRowData(row: string[] | TableAnchorRowData): row is TableAnchorRowData {
+    return "type" in row && row.type === "anchor";
+}
--- a/packages/table/README.md
+++ b/packages/table/README.md
@@ -1,0 +1,1 @@
+#  [`@fremtind/jkl-table`](https://fremtind.github.io/jokul/components/button/)
--- a/packages/table/README.md
+++ b/packages/table/README.md
@@ -1,1 +1,1 @@
-#  [`@fremtind/jkl-table`](https://fremtind.github.io/jokul/components/button/)
+#  [`@fremtind/jkl-table`](https://fremtind.github.io/jokul/components/table/)
--- a/packages/table/example/index.html
+++ b/packages/table/example/index.html
@@ -1,0 +1,64 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <meta charset="utf-8" />
+        <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+        <title>Jøkul Button</title>
+        <meta name="viewport" content="width=device-width, initial-scale=1" />
+        <link rel="stylesheet" href="index.scss" />
+    </head>
+
+    <body>
+        <table class="jkl-table">
+            <thead>
+                <tr class="jkl-table__row jkl-table__row--heading">
+                    <th class="jkl-table__heading">Dato</th>
+                    <th class="jkl-table__heading">Saksnummer</th>
+                    <th class="jkl-table__heading">Kundenummer</th>
+                    <th class="jkl-table__heading">Kundenavn</th>
+                    <th class="jkl-table__heading">Sakstype</th>
+                    <th class="jkl-table__heading">Milepæl</th>
+                    <th class="jkl-table__heading">Følger saken</th>
+                </tr>
+            </thead>
+            <tbody>
+                <tr class="jkl-table__row jkl-table__row--data-row">
+                    <td class="jkl-table__data-cell">24.02.2020</td>
+                    <td class="jkl-table__data-cell">20-1234567</td>
+                    <td class="jkl-table__data-cell">010203 99887</td>
+                    <td class="jkl-table__data-cell">Ola Nordmann</td>
+                    <td class="jkl-table__data-cell">Etterlatte</td>
+                    <td class="jkl-table__data-cell">Opprettet</td>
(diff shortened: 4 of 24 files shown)


packages/table-react/src/TableRow.tsx
```diff
@@ -0,0 +1,48 @@
...
+export interface BaseTableRowData {
+    type: TableRowDataVariants;
+}
+
+export interface TableAnchorRowData extends BaseTableRowData {
+    rowData: string[];
+    type: "anchor";
+    href: string;
+    hrefLabel: string;
+    onClickRow?: (href: string) => void;
```
lfbergee: onRowClick?

packages/table-react/src/TableRow.tsx
```diff
@@ -0,0 +1,48 @@
...
+    if (isAnchorRowData(row)) {
+        onClick = () => {
+            row.onClickRow && row.onClickRow(row.href);
+        };
+        rowModifierClasses += "jkl-table__row--anchor-row";
+    }
+
+    return (
+        <tr onClick={onClick} className={`jkl-table__row jkl-table__row--data-row ${rowModifierClasses}`}>
+            {rowData.map((data, j) => (
```
lfbergee: Same issue with index as key

packages/table/README.md
```diff
@@ -0,0 +1 @@
+#  [`@fremtind/jkl-table`](https://fremtind.github.io/jokul/components/button/)
```
lfbergee: button copypasta

packages/table/example/index.html
```diff
@@ -0,0 +1,67 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <meta charset="utf-8" />
+        <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+        <title>Jøkul Button</title>
```
lfbergee: button

portal/src/examples/TableExample.tsx
```diff
@@ -0,0 +1,120 @@
...
+            <H1>Enkel tabell</H1>
+            <Table rows={rows} columns={columns} />
+
+            <H1 className="jkl-spacing--top-4">Enkel tabell med lenker</H1>
+            <Table rows={anchorRows} columns={columns} />
+        </>
+    );
+};`;
+
+const exampleImport = `import { PrimaryButton, SecondaryButton, TertiaryButton } from "@fremtind/jkl-table-react";
```
lfbergee: Probably not right

packages/table-react/src/TableRow.tsx
Saegrov: Agree!

packages/table-react/src/TableRow.tsx
```diff
@@ -0,0 +1,48 @@
...
+export function TableRow(props: Props) {
+    const { row } = props;
+    const rowData: string[] = isAnchorRowData(row) ? row.rowData : row;
+
+    let onClick: undefined | ((evt: MouseEvent<HTMLTableRowElement>) => void) = undefined;
+    let rowModifierClasses = "";
+
+    if (isAnchorRowData(row)) {
+        onClick = () => {
+            row.onRowClick && row.onRowClick(row.href);
```
piofinn: Should this default to some function (like `document.location.assign(row.href)`)? I feel like that would be the expected behavior, and if you don't pass a function—like in the example—nothing happens when you click the row outside the first data cell.

packages/table/table.scss
```diff
@@ -0,0 +1,70 @@
...
+            &:hover {
+                background-color: white;
+                box-shadow: 0 4px 3px -3px rgba(0, 0, 0, 0.2);
+            }
+            &:active {
+                box-shadow: 0 ($bottom-border-size * 2) 0px -1px rgba(0, 0, 0, 0.4);
+            }
+        }
+
+        /* Focus-within is not supported by IE/Edge */
```
piofinn: 😥

portal/src/examples/TableExample.tsx
```diff
@@ -0,0 +1,120 @@
...
+                "Kari Nordkvinne",
+                "Etterlatte",
+                "Opprettet",
+                "Per Persen",
+            ],
+        },
+    ];
+    return (
+        <>
+            <H1>Enkel tabell</H1>
```
piofinn: Should probably be a `H3` since the example has a level 2 heading. This `H1` breaks the document flow (and looks a little imposing 😅).

portal/src/examples/TableExample.tsx
```diff
@@ -0,0 +1,120 @@
...
+            <Table rows={rows} columns={columns} />
+
+            <H1 className="jkl-spacing--top-4">Enkel tabell med lenker</H1>
```
piofinn: Same as above

packages/table-react/src/Table.tsx
```diff
@@ -0,0 +1,30 @@
...
+                <tr className="jkl-table__row">
+                    {props.columns.map((columnValue, i) => (
+                        <td className="jkl-table__heading" key={i}>
+                            {columnValue}
+                        </td>
+                    ))}
+                </tr>
+            </thead>
+            <tbody>
+                {props.rows.map((row, i) => (
```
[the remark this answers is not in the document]
Saegrov: I followed the principle of YAGNI here.

https://reactjs.org/docs/reconciliation.html

> As a last resort, you can pass an item’s index in the array as a key. This can work well if the items are never reordered, but reorders will be slow.

>Reorders can also cause issues with component state when indexes are used as keys. Component instances are updated and reused based on their key. If the key is an index, moving an item changes it. As a result, component state for things like uncontrolled inputs can get mixed up and updated in unexpected ways.

1. We don't have anyone (yet) with really big tables. Reordering a table with 10 rows doesn't really get slow even if reordering is slow.
2.  Only valid children of the table are strings and they cannot have uncontrolled state, so no Heisenbugs.

Imo we should add the ability to pass a row ID when it is needed.

portal/src/examples/TableExample.tsx
```diff
@@ -0,0 +1,120 @@
...
+};`;
+
+const exampleImport = `import { Table } from "@fremtind/jkl-table-react";
+import "@fremtind/jkl-table/table.min.css";`;
+
+const TableExample = () => (
+    <Example exampleComponents={{ Table, H1 }} exampleCode={example} type={tableType} exampleImport={exampleImport} />
+);
+
+export default TableExample;
```
piofinn: To solve the heading issue, maybe make this into two examples and place the second example beneath the readme on the component  page? We should probably take another look at the page structure altogether...

packages/table-react/src/Table.tsx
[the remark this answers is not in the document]
Saegrov: I'm very open on changing this btw @piofinn @lfbergee but as far as I can see we don't have any of the problems in the medium article. Any thoughts?

packages/table-react/src/TableRow.tsx
Saegrov: Se comment above.

portal/src/pages/components/table.tsx
```diff
@@ -0,0 +1,21 @@
...
+import readmeContents from "!raw-loader!@fremtind/jkl-table/README.md";
+// @ts-ignore
+import changelog from "!raw-loader!@fremtind/jkl-table/CHANGELOG.md";
+// @ts-ignore
+import changelogReact from "!raw-loader!@fremtind/jkl-table-react/CHANGELOG.md";
+
+export default function TablePage() {
+    return (
+        <ComponentLayout title="Tabell" mainExample={<TableExample />} packageName="table">
+            <ReactMarkdown renderers={renderer} source={readmeContents} />
```
piofinn: Per above you could add a second example here with the `Example` component directly

packages/table/table.scss
```diff
@@ -0,0 +1,70 @@
...
+    &__row {
+        box-shadow: 0 ($bottom-border-size * 2) 0px -1px rgba(0, 0, 0, 0.2);
+
+        &--anchor-row {
+            cursor: pointer;
+        }
+
+        &--data-row {
+            transition: background-color 150ms ease, box-shadow 150ms ease;
+            &:hover {
```
piofinn: Should we distinguish between normal and clickable rows somehow? It feels a bit unnatural not to be able to click on regular rows with the current style, and the changing cursor is kind of a weak distinction IMO.

packages/table/table.scss
```diff
@@ -0,0 +1,70 @@
+@import "~@fremtind/jkl-core/variables/_all.scss";
+
+$bottom-border-size: 1px;
+
+.jkl-table {
+    border-collapse: collapse;
+
+    &__anchor {
+        position: absolute;
```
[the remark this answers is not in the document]
Saegrov: Yes. Either text-selection or right-click context menu.

As it is now, we only have right-click on the first cell. So maybe we should remove it for all?

packages/table/table.scss
```diff
@@ -0,0 +1,70 @@
+@import "~@fremtind/jkl-core/variables/_all.scss";
+
+$bottom-border-size: 1px;
```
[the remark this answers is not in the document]
Saegrov: Yes.

packages/table-react/src/Table.tsx
```diff
@@ -0,0 +1,30 @@
...
+    rows: Array<TableRowData>;
+}
+
+export function Table(props: Props) {
+    return (
+        <table className="jkl-table">
+            <thead>
+                <tr className="jkl-table__row">
+                    {props.columns.map((columnValue, i) => (
+                        <td className="jkl-table__heading" key={i}>
```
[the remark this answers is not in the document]
Saegrov: Yes.

packages/table/table.scss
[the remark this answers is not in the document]
piofinn: I think removing it for all might be better.. If we do `z-index: 0` or `width: 0; height: 0;` it should still make sense to screen readers, right?

packages/table-react/src/TableRow.tsx
Saegrov: Yes you are absolutely are right and we should navigate by default.

packages/table/table.scss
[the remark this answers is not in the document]
Saegrov: We should create a `jkl-screen-reader-only` class for these cases. 
Like the one NRK (and many other) have https://github.com/nrkno/core-css/blob/31f43b7273fa28f4931b199f36f79ae2d4a9d897/lib/core-css.css#L379

packages/table/table.scss
Saegrov: I found no way of using css only to trigger focus state. Only other way is to use js and onFocus/onBlur and add focus-classes on the row. Maybe we should add this? @piofinn

packages/table/table.scss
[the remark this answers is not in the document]
Saegrov: Fixed with `jkl-sr-only`.

packages/table-react/src/TableRow.tsx
Saegrov: I also added `preventDefault` on the anchor element so that all navigation goes through the TableRow click handler.

portal/src/examples/TableExample.tsx
Saegrov: Fixed.

portal/src/examples/TableExample.tsx
Saegrov: Imo we should restructure to allow for this. And do it in a separate PR. I'll leave this as is for now.

portal/src/pages/components/table.tsx
Saegrov: See comment above.

packages/table/table.scss
piofinn: Yeah, might as well do that. Some accessibility is better than none! There is [a polyfill](https://www.npmjs.com/package/focus-within-polyfill) for focus-within that we should probably look into before we end up rewriting it ourselves...

portal/src/examples/TableExample.tsx
piofinn: We do allow for this, though! See reply below 😊

portal/src/pages/components/table.tsx
piofinn: You can insert a new `Example` component below the markdown if you want (you just have to export/import it). Something like:
```jsx
import { ComponentLayout, Changelog, Example } from "../../components";
import { Table } from "@fremtind/jkl-table-react";
import { secondExample } from "../../examples/TableExample";
// then, inside ComponentLayout
<h2 className="jkl-h3">Example with links</h2>
<Example exampleCode="secondExample" exampleComponents={{ Table }} />
```
A bit of a hassle, and sort of verbose, but it works :)

packages/table-react/src/Table.tsx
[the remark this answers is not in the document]
piofinn: Not a problem right now. The question is, do we want to have to remember to fix this when someone decides they need a sortable table with 100 rows?

packages/table/table.scss
Saegrov: I added documentation about the polyfill.

portal/src/pages/components/table.tsx
Saegrov: Fixed!

portal/src/examples/TableExample.tsx
Saegrov: Aight. I fixed this now.

packages/table-react/src/Table.tsx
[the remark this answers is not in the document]
Saegrov: Hopefully we will fix it when it becomes a problem and not before. To cite the grand dad of programming Donald Knuth:

> The real problem is that programmers have spent far too much time worrying about efficiency in the wrong places and at the wrong times; premature optimization is the root of all evil (or at least most of it) in programming.

packages/table/table.scss
Saegrov: Good catch. This was a misunderstanding between me and the designer. There should be no hover state on normal rows.

portal/src/pages/components/table.tsx
Saegrov: Fixed.

PR #373, opened 2019-09-23
Standardize form components

### LINKED ISSUE #359: Radiobutton displays square "dot" while pressed
While pressing the radio button the dot displays as a sharp square. It disappears, and the transition plays as expected, when mouse is released.

**To Reproduce**

1. Use the radio button (e.g. in the portal example)

**Expected behavior**
The dot should not be displayed while pressed, but animate in on release

### LINKED ISSUE #362: Separate value from display name in dropdown component

Alternative answers in AHV includes both value and display name. I.e. if we need to select a country, we could have a model like the following:

```
const choices = [
  { "value": 1, "name": "Norge" },
  { "value": 2, "name" "Sverige" },
]
```

In this case, we'd like to display `name` to the user, whilst we use `value` internally. In Jøkul, we have to index values by name and translate back and forth to handle changes:

```
const valuesIndexedByName = choices.reduce((acc, { value, name }) => ({
        ...acc,
        [name]: value,
    }), {});

<Dropdown
        label=""
        items={Object.keys(valuesIndexedByName)}
        onChange={name => answerQuestion(valuesIndexedByName[name])}
    />
```

Ideally, the dropdown component should support value and display value fields so that we could render the component as shown below.

```
<Dropdown
        items={choices}
        onChange={value => answerQuestion(value)}
    />
```

Additionally, we don't want to submit a label field into the component, this should be optional. In our case the label is rendered elsewhere (see #330).

**Describe alternatives you've considered**

We currently translate between text and value outside the component, it works, but it's not a good solution.

### LINKED ISSUE #366: Standardising and formalising Form-components
We have created a new structure for form components to be able to build consistent forms across our various applications.

The new form requires changes to the Input field and Dropdown components, and standardisation of the "Form group" component(?).

**New structure:**
Every form component will consist of a label + the form component itself. There will be 2 label sizes available to chose from depending on if the form component stands alone in its section or if it is in a section with multiple form components.

For sizes and references see the Figma link and ask Mikkel / Stein Olav

**Changes to inputs & dropdowns:**
Input and Dropdown components will not have animation to change label size, instead they will have a placeholder text until input or selection is made. 

**Figma link**
https://www.figma.com/file/TkbB9ANfejDSjB2u4u1OEuqM/J%C3%B8kul-komponenter-byggesone?node-id=4055%3A6026

piofinn:
## 📥 Proposed changes

Update the `Datepicker`, `Dropdown`, `FieldGroup`, `RadioButtonChoice`, `Select`, `TextField` and `TextArea` to use the new label styles and variants. Move the label styles themselves to `core`, alongside the `SupportLabel` styling.

Along with the cosmetic changes there are some functional and conceptual changes that are listed below for discussion. **Please read the entire thing before commenting/reviewing**.

Closes #366, closes #359, closes #362 

## ☑️ Submission checklist

## 💬 Further comments

Here are some changes that might need some further explanation, or are good discussion points. I've tried to sort these somewhat by component or theme.

### Label

The labels now come in three variants: A large one (default) along with `secondary` and `small` (subject to renaming) variants. These are styled with the new `.jkl-label`-classes found in `core/labels.scss`.

While specific mobile/compact versions are _not_ implemented, the font sizes and line heights of all variants are specified with `scss` functions, meaning they _will_ become smaller on small screens. I suggest that the designers take a look at the implementation as is, and suggest changes needed to match the sketches both design-wise and conceptually.

We should further discuss extracting the label to a separate component for a few reasons: Right now, there is a lot of repetition in the components due to passing down the `variant` to the label inside. Furthermore, there is the proposed use in AUW where a description follows the question/label, _before_ the input itself.

### Datepicker

Datepicker now only uses the `text-field` style, not the React component. This was done to have better control over structure in the component, specifically to place the "dropdown" portion correctly, above the support label.

This _does_ lead to some structure being duplicated between `DatePicker` and `TextField`, but removes a dependency in `DatePicker`. The alternative solution would have been to allow an error state on `TextField` without supplying an error text (so we could display the `SupportLabel` in `DatePicker` instead) , which we probably don't want.

### Dropdown

The "native" `Select` component _must_ now be used as a controlled component. This is probably the preferred use in most cases, but is worth discussing.

The `Dropdown` _can not_ be used as a controlled component. This is not due to any changes in this PR, but should probably be discussed in relation to the above. We should provide a consistent experience with both components.

These components should be renamed to `Select` and `NativeSelect`, but I wanted to do that in a separate PR since the changes made here would be hard to follow.


packages/datepicker-react/src/DatePicker.tsx
```diff
@@ -23,6 +22,7 @@ interface Props {
     className?: string;
     helpLabel?: string;
     errorLabel?: string;
+    variant?: "secondary" | "small";
```
piofinn: We might want to move this type to `core` to make it easier to sync with eventual changes to label variants

Mikkelinski:
"We should further discuss extracting the label to a separate component for a few reasons: Right now, there is a lot of repetition in the components due to passing down the variant to the label inside. Furthermore, there is the proposed use in AUW where a description follows the question/label, before the input itself."

I have no issues with separating the label as long as the design / function remains intact. I have a bigger issue with allowing AUW to have a separate style than our other inputs - we are already allowing for more variation than I believe is strictly necessary. If description text has to be below the label then that should be the standard for all going forwards, if not then AUW should also require their descriptions  to be bellow the input fields - I can't think of a reason for why we should allow this to differ between implementations.

.eslintrc.js
```diff
@@ -29,5 +29,6 @@ module.exports = {
         "@typescript-eslint/explicit-function-return-type": 0, // Explicit function return type becomes very annoying for React Function Components
         "@typescript-eslint/no-use-before-define": ["error", { functions: false }],
         "@typescript-eslint/ban-ts-ignore": 0, // We use ts-ignore for modules that don't have type definition files
+        "jsx-a11y/no-onchange": 0, // To allow the onChange method on <select> elements
```
lfbergee: Do we want do disable this for all cases? If it's just for the spesific case of select, I think we should escape it for that particular case

packages/core/labels.scss
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    &--error {
+        color: $error-color;
+    }
+}
+
+.jkl-label {
+    display: inline-block;
+    @include font-size("xl");
```
lfbergee: Since font-size and line height normally goes together in the same size, should that be default behavior, and we can override in other cases?

piofinn: Not a bad idea. For this application we should probably implement an analogue to `@font-size/@line-height` for the spacing as well.

packages/datepicker-react/src/DatePicker.tsx
```diff
@@ -112,31 +113,45 @@ export function DatePicker({
...
-                    className="jkl-datepicker__calendar"
-                >
-                    <div className="jkl-datepicker__calendar-header">
-                        <TextField label={yearLabel} type="year" className="jkl-datepicker__calendar-header--year" />
-
-                        <Select className="jkl-datepicker__calendar-header--month" label={monthLabel} items={[]} />
-                    </div>
-                    <table data-testid="jkl-datepicker-calendar" />
-                </CoreDatepicker>
+            <div className="jkl-datepicker__outer-wrapper">
```
lfbergee: outer-wrapper is a bit butter on lard right?

packages/datepicker/datepicker.scss
```diff
@@ -6,16 +6,18 @@
 $date-width: rem(46px);
 $calendar-padding: $component-spacing--xl;
 $calendar-width: ($date-width * 7) + ($calendar-padding * 2);
-$background-color: $snohvit;
-$popup-top-position: $label-height + $input-height + $bottom-padding + rem(2px);
+$background-color: white;
```
lfbergee: not our color.

packages/datepicker-react/src/DatePicker.tsx
piofinn: Hehe, yeah.. The reason is that in `Dropdown` there is a `option-wrapper` element, and I wanted to differentiate it from that. I simply used the same naming convention here.

packages/dropdown-react/example/index.tsx
```diff
@@ -18,28 +18,47 @@ const DropdownDemo = () => {
...
+                onChange={(e) => setFavoriteScene(e.target.value)}
+                placeholder="Choose your favorite"
+                variant="secondary"
+                value={favoriteScene}
             />
 
-            <Dropdown label="Favorite The Room scene" items={items} />
-            <Dropdown className="short" label="Fødselsår" items={years} />
+            <Dropdown className="jkl-spacing--top-5" label="Favorite The Room scene" items={items} />
+            <Dropdown className="short jkl-spacing--top-3" label="Fødselsår" items={years} />
```
lfbergee: short?

packages/datepicker/datepicker.scss
piofinn: Pure white _is_ a support color, but it needs to be implemented as a variable, along with the new red/blue support colors. I'll make a issue!

packages/dropdown-react/src/Dropdown.tsx
```diff
@@ -73,72 +80,55 @@ export const Dropdown = ({
...
-                            </button>
-                        </li>
-                    ))}
-                </ul>
-            </CoreToggle>
-
-            <span className="jkl-dropdown__chevron" />
-            <SupportLabel helpLabel={helpLabel} errorLabel={errorLabel} />
-        </div>
+        <>
```
lfbergee: No need for the fragment

packages/dropdown-react/src/Select.tsx
```diff
@@ -20,23 +23,43 @@ export function Select({
...
     helpLabel,
     errorLabel,
+    variant,
+    placeholder,
+    value,
     ...rest
 }: Props) {
-    const handleChange = (event: React.ChangeEvent<HTMLSelectElement>) => onChange && onChange(event);
+    // Set value to value given, or to first item if no value or placeholder is given:
+    value = value ? value : placeholder ? "" : items[0];
```
lfbergee: Nested turnary... I do not like... This should not pass linting. The cognative load of reading that line makes me sleepy.

packages/dropdown/dropdown.scss
```diff
@@ -139,40 +112,50 @@ $chevron-weight: rem(1px);
...
-        background-color: white;
-        list-style-type: none;
         position: absolute;
         left: 0;
         right: 0;
-        top: $dropdown-total-height;
+        top: 100%;
+        z-index: $z-index--dropdown;
+
+        background-color: white;
```
lfbergee: not our color

packages/dropdown/dropdown.scss
```diff
@@ -139,40 +112,50 @@ $chevron-weight: rem(1px);
...
         &:hover {
-            @include nojump-bold();
-            border-color: $border-color;
+            background-color: $svart;
+            color: white;
         }
 
         html:not([data-mousenavigation]) &:focus {
-            border-color: $border-color--focus;
+            background-color: white;
```
lfbergee: not our color

packages/field-group-react/src/FieldGroup.tsx
```diff
@@ -7,12 +7,17 @@ interface Props {
     className?: string;
     helpLabel?: string;
     errorLabel?: string;
+    variant?: "secondary" | "small";
```
lfbergee: Yeah, we should have this a single place

packages/dropdown-react/example/index.tsx
piofinn: I made a support class for the example (long time ago) to show that width can be adjusted by custom `className`s. I agree that it might be a bit confusing…

packages/dropdown-react/src/Select.tsx
piofinn: I'll refactor it 😅

packages/field-group-react/src/FieldGroup.tsx
piofinn: Any ideas on where/how? Maybe a `data-types.ts` file in `jkl-core/src/utils`?

packages/dropdown-react/src/Select.tsx
lfbergee: I think we should add a lint rule to avoid this in the future

packages/dropdown-react/src/Dropdown.tsx
```diff
@@ -49,13 +51,22 @@ export const Dropdown = ({
...
     const hasSelectedValue = typeof selectedValue !== "undefined";
     const listRef = useListNavigation();
+    const componentClassName = "jkl-dropdown".concat(
+        inline ? ` jkl-dropdown--inline` : "",
+        dropdownIsShown ? ` jkl-dropdown--open` : "",
+        !hasSelectedValue ? ` jkl-dropdown--no-value` : "",
+        !!errorLabel ? ` jkl-dropdown--invalid` : "",
+        className ? ` ${className}` : "",
+    );
+    const labelClassName = "jkl-label".concat(variant ? ` jkl-label--${variant}` : "");
```
Saegrov: I'm torn. I really like the `concat`-style for `componentClassName`, but not for `labelClassName`.

packages/dropdown-react/src/Dropdown.tsx
```diff
@@ -69,76 +80,63 @@ export const Dropdown = ({
...
+                    aria-activedescendant={hasSelectedValue && `${listId}__${lower(selectedValue)}`}
+                >
+                    <ul
+                        className="jkl-dropdown__option-wrapper"
+                        data-testid="jkl-dropdown__option-wrapper"
+                        tabIndex={-1}
+                        ref={listRef}
+                    >
+                        {items.map((item) => {
+                            item = getSelectValuePairFrom(item);
```
Saegrov: Personally I'm not a fan of reassigning function parameters but I'd rather create a new variabel. Specially here, since the type changes (for some cases) and its har do know what exactly `item` is.

packages/dropdown-react/src/Select.tsx
```diff
@@ -20,23 +27,53 @@ export function Select({
     inline = false,
     helpLabel,
     errorLabel,
+    variant,
+    placeholder,
+    value,
```
Saegrov: I do not know if I like it, but you could set value to a default parameter:
```
value = placeholder || getSelectValuePairFrom(items[0]).value
```

And remove the `if(value)`-stuff.

packages/dropdown-react/src/Select.tsx
```diff
@@ -20,23 +27,53 @@ export function Select({
...
+                {...rest}
+            >
+                {placeholder && value === "" && (
+                    <option disabled value="">
+                        {placeholder}
                     </option>
-                ))}
+                )}
+                {items.map((item) => {
+                    item = typeof item === "string" ? { value: item, label: item } : item;
```
Saegrov: Use `getSelectValuePairFrom`?

Saegrov: You can also use it in a map chain: 
```
                {items.map(getSelectValuePairFrom).map((item) => (
                    <option data-testid="jkl-dropdown__option" key={item.value} value={item.value}>
                        {item.label}
                    </option>
                ))}
```

packages/dropdown-react/src/Dropdown.tsx
Saegrov: You can also use it in a map chain:           
```
     {items.map(getSelectValuePairFrom).map((item) => (
                    <option data-testid="jkl-dropdown__option" key={item.value} value={item.value}>
                        {item.label}
                    </option>
                ))}
```

packages/dropdown-react/src/SelectValuePair.ts
```diff
@@ -0,0 +1,8 @@
+export interface SelectValuePair {
+    value: string;
+    label: string;
+}
+
+export function getSelectValuePairFrom(item: string | SelectValuePair) {
```
Saegrov: Might consider adding a return type for clarity.

packages/radio-button-react/example/index.tsx
```diff
@@ -13,17 +15,37 @@ const Demo = () => {
     const [selectedValue, setSelectedValue] = React.useState();
     const [inline, setInline] = useState(false);
     const [invalid, setInvalid] = useState("");
+    const [variant, setVariant] = useState<labelVariant | undefined>();
+    const typecheckAndSetVariant = (e: React.ChangeEvent<HTMLSelectElement>) => {
+        const val = e.target.value;
+        if (val === ("secondary" || "small")) {
```
Saegrov: This will never be true for `small`, will it?

packages/text-input-react/example/index.tsx
```diff
@@ -16,28 +16,44 @@ const TextFieldDemo = () => {
...
+                    onChange={handleChange}
+                    placeholder={"Norsk"}
+                    variant="small"
+                    helpLabel="La oss se..."
+                />
+                <TextField label="Fornavn" value={"Per"} onChange={handleChange} readOnly variant="small" />
             </div>
 
-            <TextField label="Telefon" type="tel" value={value} onChange={handleChange} />
+            <TextField label="Pelefon" type="tel" value={value} onChange={handleChange} placeholder="999 99 999" />
```
Saegrov: Pelefon?

packages/core/normalize.scss
```diff
@@ -42,24 +42,6 @@ h1 {
...
-h1,
-h2,
-h3,
-h4,
-h5,
-h6,
-p {
-    margin: 0;
-}
-
```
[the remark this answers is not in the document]
piofinn: Yeah, I agree. Maybe in a different PR, though (this PR is already quite bloated)? I can make an issue!

packages/text-input-react/example/index.tsx
piofinn: Haha! We were testing visual alignment and wanted a left-aligned character. Forgot to change it back 😅 Either way, we need better and domain relevant examples of most components…

packages/core/src/types/data-types.ts
```diff
@@ -0,0 +1 @@
+export type labelVariant = "secondary" | "small";
```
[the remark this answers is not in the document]
piofinn: Yep, I  see that this is recommended by the TS team as well

packages/core/src/types/index.ts
```diff
@@ -0,0 +1,5 @@
+import { labelVariant } from "./data-types";
+
+// Workaround for Babels --isolatedModules setting
+// see https://github.com/facebook/create-react-app/issues/6054
+export declare type labelVariant = labelVariant;
```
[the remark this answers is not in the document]
piofinn: Directly in root `index.ts` is a good idea!

packages/core/labels.scss
```diff
@@ -0,0 +1,42 @@
...
+.jkl-label {
+    display: inline-block;
+    margin-bottom: $component-spacing--xl;
+    margin-left: rem(-1px); // adjust visual alignment
+    @include text-sizing("xl");
+    @include small-device {
+        margin-bottom: $component-spacing--large;
+    }
+
+    &--secondary {
```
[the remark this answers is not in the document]
piofinn: Agreed. This is definitely subject to renaming! Will discuss with design and find suitable naming convention

packages/datepicker-react/src/DatePicker.tsx
```diff
@@ -112,31 +114,45 @@ export function DatePicker({
...
+                <div hidden={datepickerHidden}>
+                    <CoreDatepicker
+                        timestamp={date ? date.getTime() : undefined}
+                        months={months}
+                        days={days}
+                        onDatepickerClickDay={onClickCalendarDay}
+                        className="jkl-datepicker__calendar"
+                    >
+                        <div className="jkl-datepicker__calendar-header">
+                            <label className={`jkl-text-field jkl-datepicker__calendar-header--year`}>
```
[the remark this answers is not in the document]
piofinn: Yeah, these have annoyed me before. I'll fix it here while I remeber it!

packages/radio-button-react/example/index.tsx
piofinn: I chould just use the type here (it is a typecheck after all) 😅

packages/core/labels.scss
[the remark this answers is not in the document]
piofinn: Maybe even have all three variants named and update the type accordingly, and have one be standard…

packages/dropdown-react/src/Dropdown.tsx
piofinn: Haha! I agree that it works better with multiple variables, though. How about just
```tsx
const labelVariantClass = variant ? ` jkl-label--${variant}` : "";
…
<label className={"jkl-label" + variantClass}>¡Hola!</label>
```
for the labels?

packages/core/normalize.scss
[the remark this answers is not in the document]
Saegrov: Aye.

packages/datepicker-react/src/DatePicker.tsx
```diff
@@ -112,31 +114,45 @@ export function DatePicker({
...
+                                <span className="jkl-label jkl-label--small">{yearLabel}</span>
+                                <input type="year" className={`jkl-text-field__input`} />
+                            </label>
+
+                            <Select
+                                variant="small"
+                                className="jkl-datepicker__calendar-header--month"
+                                label={monthLabel}
+                                items={[]}
```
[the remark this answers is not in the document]
piofinn: I'll just refactor out the entire `<Select>` component and get rid of the dependency (and this issue), like I did with the text inputs.

packages/dropdown-react/src/Select.tsx
piofinn: I discovered we also need a check for if there are any items, so it gets a bit complicated for inlining, imo.

PR #375, opened 2019-09-24
Dropdown value pairs

### LINKED ISSUE #362: Separate value from display name in dropdown component

piofinn:
## 📥 Proposed changes

Let `Dropdown` and `Select` accept value/label pairs in addition to `string`s as items. This matches native behaviour better, and closes #362 .

## ☑️ Submission checklist

## 💬 Further comments

The form of the value/label pairs is defined as a type (`SelectValuePair`) in `SelectValuePair.ts` and can be imported from there. Both `Dropdown` and `Select` accepts an array of _either_ `string`s or `SelectValuePair`s as the `items` prop.

```ts
type SelectValuePair = {
    value: string;
    label: string;
};
```
```ts
interface Props {
    ...
    items: Array<string | SelectValuePair>;
}
```

### DIFF
--- a/packages/dropdown-react/src/Dropdown.tsx
+++ b/packages/dropdown-react/src/Dropdown.tsx
@@ -5,15 +5,11 @@ import nanoid from "nanoid";
 import { SupportLabel } from "@fremtind/jkl-typography-react";
 import { labelVariant } from "@fremtind/jkl-core";
 import { useListNavigation } from "./useListNavigation";
-
-export type SelectValue = {
-    value: string;
-    label: string;
-};
+import { SelectValuePair, getSelectValuePairFrom } from "./SelectValuePair";
 
 interface Props {
     label: string;
-    items: Array<string | SelectValue>;
+    items: Array<string | SelectValuePair>;
     inline?: boolean;
     defaultPrompt?: string;
     className?: string;
@@ -43,10 +39,6 @@ function focusSelected(listEl: HTMLElement, listId: string, selected: string | u
         focusedItem = listEl.querySelector('[role="option"]');
     }
     focusedItem && focusedItem.focus();
-}
-
-export function makeSelectValueFrom(item: string | SelectValue) {
-    return typeof item === "string" ? { value: item, label: item } : item;
 }
 
 export function Dropdown({
@@ -119,7 +111,7 @@ export function Dropdown({
                         ref={listRef}
                     >
                         {items.map((item) => {
-                            item = makeSelectValueFrom(item);
+                            item = getSelectValuePairFrom(item);
                             return (
                                 <li key={item.value}>
                                     <button
--- a/packages/dropdown-react/src/Dropdown.tsx
+++ b/packages/dropdown-react/src/Dropdown.tsx
@@ -86,7 +86,7 @@ export function Dropdown({
         e.target.button.focus();
         e.target.value = e.detail;
         setSelectedValue(e.detail.textContent);
-        onChange && onChange(e.detail.textContent);
+        onChange && onChange(selectedValue || "");
     }
 
     return (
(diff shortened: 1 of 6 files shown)


packages/dropdown-react/src/Dropdown.tsx
```diff
@@ -5,9 +5,14 @@ import nanoid from "nanoid";
 import { SupportLabel } from "@fremtind/jkl-typography-react";
 import { useListNavigation } from "./useListNavigation";
 
+export type SelectValue = {
```
piofinn: Should this be exported from `index.ts` as well?

piofinn:
@stianlik, does this solve your issue? I can't request your review since you're not in the organization, but please take a look! 😊

stianlik:
@piofinn Yes, this is exactly what we need, thanks :)

packages/dropdown-react/src/Dropdown.tsx
lfbergee: Not a huge fan of the name. SelectValuePair or SelectObject maybe

packages/dropdown-react/src/Dropdown.tsx
```diff
@@ -78,7 +87,7 @@ export function Dropdown({
         e.target.button.focus();
         e.target.value = e.detail;
         setSelectedValue(e.detail.textContent);
-        onChange && onChange(e.detail.textContent);
+        onChange && onChange(selectedValue || "");
```
lfbergee: Can we end up with a race condition here? Perhaps safer to use e.detail.textContent.

```
const nextSelectedValue = e.detail.textContent;
setSelectedValue(nextSelectedValue)
onChange(nextSelectedValue || "")
```

packages/dropdown-react/src/Dropdown.tsx
```diff
@@ -40,6 +45,10 @@ function focusSelected(listEl: HTMLElement, listId: string, selected: string | u
     focusedItem && focusedItem.focus();
 }
 
+export function makeSelectValueFrom(item: string | SelectValue) {
```
lfbergee: In my mind `make` is a keyword for a function that returns a function. I'd prefer `getSelectValue()` here

lfbergee: And since both uses this, i'd prefer it to be split out to it's own file or utility file

lfbergee: And write a small litte test for it :dancer:

packages/dropdown-react/src/Dropdown.tsx
piofinn: I was going to say that `e.detail.textValue` is the wrong value here, which it is, but then I saw that it is used as it is now as well 😑 We want to return the `value` from the value pair, but the text content is the `label`. I'll take a look at getting this right.

PR #402, opened 2019-10-03
Expand table

glemvik:
## 📥 Proposed changes

This expands the current table component.
In addition to ordinary rows and rows that act as links, the component now also supports rows that may be expanded to show more content (accordions).

## ☑️ Submission checklist


packages/table-react/src/Table.tsx
```diff
@@ -1,31 +1,92 @@
...
 
-export function Table(props: Props) {
+export function Table({ columns, rows, className, accordion }: Props) {
+    const componentClassName = "jkl-table".concat(
+        accordion ? ` jkl-table--accordion` : "",
+        className ? ` ${className}` : "",
+    );
+
+    const createAppropriateTableRow = (row: TableRowData, index: number) => {
+        // TODO: Good/bad practice å ha keys her?
```
lfbergee: No need for key here, inside this scope there is no reorder reconcilliation conflict. The key should go where the map is. And index is not too good of a solution. Should push this out to client developer to supply a uniqe key for each row/col.

packages/table-react/src/TableRowAnchor.tsx
```diff
@@ -0,0 +1,29 @@
...
+    row: TableRowAnchorData;
+}
+
+export function TableRowAnchor({ row }: PropTypes) {
+
+    const onClick = () => {
+        if (row.onRowClick) {
+            row.onRowClick(row.href);
+        } else {
+            window.location.href = row.href;
```
lfbergee: I always get a bit anxious when i see window.stuff in libary code. At least we should check if window is defined before firing off this. And in general modifying the history object is a can of worms.

piofinn:
I'm closing this now, as it has been available for conversation for a while now without any activity :)

Issue #323, opened 2019-08-14
Components: Accordion


Steinop:
https://www.figma.com/file/Wh8Gfnesn25shH31sKyIaT/J%C3%B8kul-bibliotek?node-id=47%3A1

Steinop:
Accordion updated with new design and different sizes - both as a regular page component and as full width Accordion sections on a page.

Animated version:
![Fremtind_Accordion v 1](https://user-images.githubusercontent.com/51952891/63027324-09876a80-bead-11e9-85eb-9b11a0c8db12.gif)

Figma (Final design will be updated to Figma Bibliotek at a later stage):
https://www.figma.com/file/TkbB9ANfejDSjB2u4u1OEuqM/J%C3%B8kul-components?node-id=218%3A348

stianlik:
I'm working on this.

Mikkelinski:
Further work handled by #394

PR #404, opened 2019-10-03
feat(accordion): implement new design for regular page component

stianlik:
affects: @fremtind/jkl-accordion

## 📥 Proposed changes

See #323 . This PR contains style changes to display regular page component.

## ☑️ Submission checklist

-   [not appropriate] I have added tests that prove my fix is effective or that my feature works
-   [not appropriate] I have added necessary documentation (if appropriate)

## 💬 Further comments

I chose not to start on a full width component, because that will likely require changes in jkl-accordion-react.

### DIFF
--- a/packages/accordion/accordion.scss
+++ b/packages/accordion/accordion.scss
@@ -1,71 +1,58 @@
 @import "~@fremtind/jkl-core/variables/_all.scss";
 @import "~@fremtind/jkl-core/mixins/_all.scss";
 
-$border-color: $svart;
-$border-width: rem(1px);
-
-// chevron
-$chevron-size: rem(20px);
-$chevron-weight: rem(1px);
-$chevron-weight--hover: rem(2px);
-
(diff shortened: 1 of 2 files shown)


packages/accordion/accordion.scss
```diff
@@ -1,71 +1,56 @@
...
-}
+    &:last-child {
+        margin-bottom: 0;
+    }
 
-.jkl-accordion-item {
-    border-bottom: $border-width solid $border-color;
-    padding: $item-padding-top-bottom 0;
+    &:active {
+        transform: scale(0.99);
```
piofinn: Maybe add a slight transition here? It feels a little harsh compared to the other changes (expand, color change)

stianlik: Good catch, I see it now. Updated PR with transition. Also added `transition-property` to content wrapper to narrow down the transition scope.

PR #413, opened 2019-10-04
feat(loader): add aria-label and title to loader component

### LINKED ISSUE #405: Loader-react is not accessible
We need some ARIA values, and the option to send in a loading text for screen readers, on this component.

See https://stackoverflow.com/questions/38704467/how-to-label-a-loading-animation-for-wai-aria for a good solution

piofinn:
affects: @fremtind/jkl-loader-react, @fremtind/jkl-loader

## 📥 Proposed changes

Add a `title` and `aria-label` to the loader to make it accessible. The text can be set via an optional
prop, and has a default value. Also add `role` and `aria-busy` attributes for better accessibility compliance.

Closes #405 

## ☑️ Submission checklist


packages/loader-react/src/Loader.tsx
```diff
@@ -1,16 +1,24 @@
 import React from "react";
 interface Props {
+    textDescription?: string;
     negative?: boolean;
     inline?: boolean;
     className?: string;
 }
-export const Loader = ({ negative = false, inline = false, className }: Props) => {
+export const Loader = ({ textDescription = "Laster inn", negative = false, inline = false, className }: Props) => {
```
piofinn: An option to consider is making the prop mandatory, with no default, to force devs/designers to think about what the loader represents.

lfbergee: Yeah, I like that. Also, we probably should have a pattern on how to use loaders. I think we should try to avoid loaders, unless the wait time is more than say 500 ms. And then the loader should show for at least 500 ms.

PR #414, opened 2019-10-04
Accordion item title

glemvik:
## 📥 Proposed changes

This pull request contains changes that will allow the title of accordion items to be React components, as well as strings.
Allowing these React components will let users style accordion item titles according to their needs.

## ☑️ Submission checklist

### DIFF
--- a/packages/accordion-react/src/AccordionItem.tsx
+++ b/packages/accordion-react/src/AccordionItem.tsx
@@ -6,7 +6,7 @@ import React, { ReactNode, useState } from "react";
 import { useAnimatedHeight } from "@fremtind/jkl-react-hooks";
 
 interface Props {
-    title: string;
+    title: string | ReactNode;
     children: ReactNode;
     startExpanded?: boolean;
 }
--- a/packages/accordion/accordion.scss
+++ b/packages/accordion/accordion.scss
@@ -39,6 +39,7 @@ $title-padding-bottom: $title-padding - $item-padding-top-bottom - $border-width
         padding: $title-padding-top $title-padding $title-padding-bottom;
         text-align: left;
         width: 100%;
+        display: flex;
 
         /* Remove button styles */
         outline: none;
@@ -72,6 +73,7 @@ $title-padding-bottom: $title-padding - $item-padding-top-bottom - $border-width
         @include font-size("small");
         @include line-height("small");
         padding-right: 2rem;
+        width: inherit;
     }
 
     &__title-icon {


packages/accordion-react/src/AccordionItem.tsx
```diff
@@ -6,7 +6,7 @@ import React, { ReactNode, useState } from "react";
 import { useAnimatedHeight } from "@fremtind/jkl-react-hooks";
 
 interface Props {
-    title: string;
+    title: string | ReactNode;
```
piofinn: No need for both, since `string` is a part of the `ReactNode` type. However it might be worth keeping it just for clarity?

glemvik: Agree that it is nice to keep for clarity

glemvik:
> There's an updated version of accordion on its way in #404, so I'll defer a proper review until you've had the chance to look at that/pull it in blush

Makes sense :+1:  Pull request #404 is now merged into this pull request, without conflicts or problems.

packages/accordion/accordion.scss
```diff
@@ -34,6 +34,7 @@
         padding: $title-padding;
         text-align: left;
         width: 100%;
+        display: flex;
```
piofinn: Is this for distributing content inside the title? In that case it should probably be `display: flex` on the `.jkl-accordion__title-text` class instead (which would then not need `width: inherit`).

packages/accordion-react/src/AccordionItem.tsx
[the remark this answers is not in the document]
stianlik: I looked into this issue when I worked on design changes in #404 and sketched out some ideas for a more flexible component. See the example below (usage at the bottom).

The idea is to implement a base component for accordion items and extend this component for each variant needed. We provide a few defaults in Jøkul, and allow for other applications to implement their own with some basic building blocks (AccordionItemBase, AccordionIcon, ...).

I modified the item structure slightly by using header and body wrappers in the accordion item base component to make it easier to implement a full width design.

```JSX
//
// Suggested base components in Jøkul
//

// title and content are react components
// wrap header and body to accomodate new full width design with centered content
// this requires very little changes in style sheet and will be backwards compatible
const AccordionItemBase = ({ title, content }) => (
    <div class="jkl-accordion__item">
        <header class="jkl-accordion__header" role="button" tabindex="1">
            {title}
        </header>
        <div class="jkl-accordion-body">{body}</div>
    </div>
);

// title is type string
// Jøkul can provide a few variants (e.l. h2 and h3 as described in design suggestion for full width)
const AccordionItem = ({ title, content }) => (
    <AccordionItemBase title={<AccordionTitleButton>{title}</AccordionTitleButton>}>
        <AccordtionContent>{content}</AccordtionContent>
    </AccordionItemBase>
);

// May allow element="div|span|whatever..": if !button add aria-role="button" and tabIndex={0}
const AccordionTitleButton = ({ element }) => (
    <button class="jkl-accordion__title">
        <span class="jkl-accordion__title-text"></span>
        <AccordionIcon />
    </button>
);

const AccordionContent = ({ children }) => <div class="jkl-accordion-content">{children}</div>;

const AccordionIcon = ({ mayNeedSizeOrSomething }) => <span class="jkl-accordion__tile-icon"></span>;

//
// Custom components for other applications
//

const FancyTitle = ({ element }) => (
    <h2 class="jkl-accordion__title" role="button" tabindex={0}> {/* no button element here, problem solved! */}
        <span class="jkl-accordion__title-text"></span>
        <AccordionIcon />
    </h2>
);

const FancyItem = ({ title, content }) => (
    <AccordionItemBase title={<FancyTitle>{title}</FancyTitle>}>
        <AccordtionContent>{content}</AccordtionContent>
    </AccordionItemBase>
);

//
// Usage
//

const SomeContent = () => (
    <>
        <H1>Custom design</H1>
        <P>
              No need to pass components in to the title when we
              have a custom item component
        </P>
     
        <Accordion>
            <FancyItem title="Fancy title">And content</FancyItem>
            <FancyItem title="Another fancy title">And content</FancyItem>
        </Accordion>

        <H1>Old design</H1>
        <Accordion>
            <AccordionItem title="Same old title">And content</AccordionItem>
            <AccordionItem title="Same old title number two">And content</AccordionItem>
        </Accordion>

        <H1>Full width h2 design using props (not included in definitions above)</H1>
        <Accordion>
            <AccordionItem fullWidth variant="h2" title="Heading title">
                And content
            </AccordionItem>
            <AccordionItem fullWidth variant="h2" title="Heading title two">
                And content
            </AccordionItem>
        </Accordion>

        <H1>Full width h2 design using components (better extensibility than props version)</H1>
        <Accordion>
            <AccordionItemFullWidth variant="h2" title="Heading title">
                And content
            </AccordionItemFullWidth>
            <AccordionItemFullWidth variatn="h3" title="Heading title two">
                And content
            </AccordionItemFullWidth>
        </Accordion>
    </>
);
```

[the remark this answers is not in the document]
piofinn: Wow, you've put some effort into this @stianlik! I like the concept/architecture a lot, but this is starting to feel like a new component… Is it really an accordion if both the header and the content can be anything? It might very well be, but it feels like such a conceptual change that we should discuss it.

I guess this was always the case with this PR, but your example implementation drove the point through.

PR #417, opened 2019-10-04
feat(datepicker): implement new version and styles

piofinn:
affects: @fremtind/jkl-core, @fremtind/jkl-datepicker-react, @fremtind/jkl-datepicker

## 📥 Proposed changes

Implement the new, compact version of the datepicker, with the option of using the old version using the `extended` prop.

## ☑️ Submission checklist

### DIFF
--- a/packages/datepicker/datepicker.scss
+++ b/packages/datepicker/datepicker.scss
@@ -121,7 +121,7 @@ $bottom-margin: $component-spacing--large;
             text-transform: capitalize;
 
             .jkl-datepicker--extended & {
-                display: none;
+                @include screenreader-only();
             }
         }
         th {
--- a/packages/datepicker/datepicker.scss
+++ b/packages/datepicker/datepicker.scss
@@ -1,41 +1,90 @@
 @import "~@fremtind/jkl-core/variables/_all.scss";
+@import "~@fremtind/jkl-core/mixins/_all.scss";
 @import "~@fremtind/jkl-core/_functions.scss";
 @import "~@fremtind/jkl-text-input/text-input.scss";
 @import "~@fremtind/jkl-dropdown/dropdown.scss";
 
-$date-width: rem(46px);
+$date-width: $line-height-4;
+$date-spacing: rem(4px);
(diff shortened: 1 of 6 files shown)


packages/datepicker/datepicker.scss
```diff
@@ -1,41 +1,98 @@
...
 @import "~@fremtind/jkl-core/_functions.scss";
 @import "~@fremtind/jkl-text-input/text-input.scss";
 @import "~@fremtind/jkl-dropdown/dropdown.scss";
 
-$date-width: rem(46px);
+$date-width: $line-height-4;
+$date-spacing: rem(4px);
 $calendar-padding: $component-spacing--xl;
-$calendar-width: ($date-width * 7) + ($calendar-padding * 2);
+$calendar-width: ($date-width * 7) + ($date-spacing * 6);
```
lfbergee: This can be tight on a mobile screen, 304px as set width. Can we make this more dynamic?

packages/datepicker/datepicker.scss
```diff
@@ -1,41 +1,98 @@
...
 $background-color: $helhvit;
 $bottom-margin: $component-spacing--large;
 
 .jkl-datepicker {
     position: relative;
-    width: $calendar-width;
+    width: $calendar-width + ($calendar-padding * 2);
```
lfbergee: 352px... So yeah, we do need to figure something out here, as this will not work on most mobile devices. I guess you can argue we should use native datapicker for mobile, but then I think we need to handle that in Jøkul, not defer to client development.

Can we be more dynamic with the width property?

lfbergee: I guess it has been a problemn all along, still a problem though.. But I wont block this PR for that.

piofinn: Mikkel has a suggestion where we set the side padding to `8px` on small screens (under 360px?), which will make the total with 320px. However, we also need to make sure it positions itself edge to edge so we avoid side-scrolling.

It feels like a good point to have under the "v1.0.0" issue for this component!

PR #419, opened 2019-10-07
Max length text field

### LINKED ISSUE #418: Missing max length for text field
The textfield should have a property for specifying maxLength.

The same as a normal `input`-element where maxLength prohibits you from typing in more letters than maxLength.

frisol97:
## 📥 Proposed changes

Added max length to input-field. Closes #418

## ☑️ Submission checklist


frisol97:
> I think you're requesting a bit more than you intended. Either rebase on master, or create a new branch from master and cherry pick in your commit.

> Hehe I know, should be fixed now

packages/text-input-react/src/TextField.test.tsx
```diff
@@ -24,4 +24,10 @@ describe("TextField", () => {
         const component = getByTestId("jkl-text-field");
         expect(component).toHaveClass("test-class");
     });
+
+    it("has the max-length given", () => {
+        const { getByLabelText } = render(<TextField label="testing" maxLength={10} />);
+
+        expect(getByLabelText("testing")).toHaveAttribute("max-length", 10);
```
piofinn: I remembered wrong about the attribute name. It is `maxlength`, and the returned value is a `string`. Fix those, and it works 😊

packages/text-input-react/example/index.tsx
```diff
@@ -23,6 +23,7 @@ const TextFieldDemo = () => {
                     placeholder={"Norsk"}
```
Saegrov: I'm not sure if it is a good idea, but since we use the spread operator `{...rest}` we can remove many fields:

```
    return (
        <label data-testid="jkl-text-field" className={componentClassName}>
            <span className={labelClassName}>{label}</span>
            <input aria-invalid={!!errorLabel} className={`jkl-text-field__input`} {...rest} />
            <SupportLabel helpLabel={helpLabel} errorLabel={errorLabel} />
        </label>
    );

```

Just throwing it out there.
```

File: packages/text-input-react/src/TextField.test.tsx [hunk 3]
```diff
@@ -24,4 +24,10 @@ describe("TextField", () => {
         const component = getByTestId("jkl-text-field");
         expect(component).toHaveClass("test-class");
     });
+
+    it("has the max-length given", () => {
```
Saegrov: Yay for writing tests!

PR #421, opened 2019-10-08
Remove the need for @ts-ignore when using raw-loaders in portal

Saegrov:
## 📥 Proposed changes

Remove the need for @ts-ignore when using raw-loaders in portal.

## ☑️ Submission checklist

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -3,3 +3,5 @@ dist/
 public/
 .cache/
 jest/
+tsconfig.json
+tsconfig-for-declarations.json
--- a/portal/raw-loader.ts
+++ b/portal/raw-loader.ts
@@ -1,0 +1,4 @@
+declare module "!raw-loader*" {
+    const content: string;
+    export default content;
+}
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -3,14 +3,14 @@
         /* Basic Options */
         "target": "esnext"                        /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017','ES2018' or 'ESNEXT'. */,
         "noEmit": true                            /* Do not emit outputs. We use babel as our compiler. */,
-        "isolatedModules": true                   /* Transpile each file as a separate module (similar to 'ts.transpileModule'). This is how babel operates */,
+        "isolatedModules": false                  /* Transpile each file as a separate module (similar to 'ts.transpileModule'). This is how babel operates */,
         "lib": [                                  /* Specify library files to be included in the compilation. */
             "dom"                                 /* We have a browser environment */
         ],
         "module": "esnext"                        /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */,
         // "allowJs": true,                       /* Allow javascript files to be compiled. */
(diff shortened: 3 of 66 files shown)


.prettierignore
```diff
@@ -2,4 +2,6 @@ build/
 dist/
 public/
 .cache/
-jest/
\ No newline at end of file
+jest/
+tsconfig.json
```
piofinn: Haha, love the fact that you had to _disable_ Prettier to get nice formatting here 😆

portal/raw-loader.ts
```diff
@@ -0,0 +1,4 @@
+declare module "!raw-loader*" {
+    const content: string;
+    export default content;
+}
```
piofinn: This was deceptively simple 🤔 Well done!

Saegrov: Using this feature requires `isolatedModules` to be turned off (which I did).

I'm having 2nd doubts if this is a good idea.

tsconfig.json
```diff
@@ -1,17 +1,16 @@
...
         /* Basic Options */
-        "target": "esnext" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017','ES2018' or 'ESNEXT'. */,
-        "noEmit": true /* Do not emit outputs. We use babel as our compiler. */,
-        "isolatedModules": true /* Transpile each file as a separate module (similar to 'ts.transpileModule'). This is how babel operates */,
-        "lib": [
-            /* Specify library files to be included in the compilation. */
-            "dom" /* We have a browser environment */
+        "target": "esnext"                        /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017','ES2018' or 'ESNEXT'. */,
+        "noEmit": true                            /* Do not emit outputs. We use babel as our compiler. */,
+        "isolatedModules": false                  /* Transpile each file as a separate module (similar to 'ts.transpileModule'). This is how babel operates */,
```
lfbergee: I guess the comment no longer applies to the setting, right?

And this now opens up the possibility of the typecheck passing, and the compiling with babel to fail. If I understood correctly. Thats not great, but, might not be a huge issue as we build the project on the CI server, so it will not actually go into master.

portal/raw-loader.ts
lfbergee: It's prettier and easier to use. I like it. I think i like it more than the trade off... but maybe it should be documented somewhere

piofinn: We will be rewriting the portal pages soon™, so we can turn `isolatedModules` back on when we redo the example system anyway. Then again, we could also argue that we can just have the `@ts-ignore`s in there until then instead 🤷‍♂️

lfbergee: But will the rewrite remove the need for raw import? I think we should keep showing the definition files in the New And Improved Portal, and yeah we can do that some other way, but for now that is by raw import. The rest should be rather easy to get rid of though.

tsconfig.json
Saegrov: You're right. I'll change it.

portal/raw-loader.ts
piofinn: Yeah, you're right that it's not a given!

PR #425, opened 2019-10-10
fix(table-react): add customized click event on row

frisol97:
affects: @fremtind/jkl-table-react

## 📥 Proposed changes

Added customRowCilck to enhance the flexibility on the table component. This is a proposal to evaluate a customized click-row-approach and remove the current onRowClick.

## ☑️ Submission checklist

## 💬 Further comments

We need feedback.


packages/table-react/src/TableRow.tsx
```diff
@@ -28,13 +29,19 @@ export function TableRow(props: Props) {
...
-        onClick = () => {
-            if (row.onRowClick) {
-                row.onRowClick(row.href);
-            } else {
-                window.location.href = row.href;
-            }
-        };
+        if (row.customRowClick !== undefined) {
+            onClick = () => {
+                row.customRowClick && row.customRowClick();
```
Saegrov: This is exactly the same logic as `row.onRowClick`. Why can't `onRowClick` be used?

packages/table-react/example/index.tsx
```diff
@@ -21,12 +21,27 @@ const rows: Array<string[] | TableAnchorRowData> = [
...
+        rowData: [
+            "24.12.2020",
+            "20-2028732",
+            "010203 99887",
+            "Eduardo Nordkvinne Nordmann",
+            "Etterlatte",
+            "Opprettet",
+            "Freja Forsikringstaker",
+        ],
+        customRowClick: () => alert("Customized on row click"),
```
Saegrov: Why not use `onRowClick`? It should do exactly the same thing.

packages/table-react/src/TableRow.tsx
```diff
@@ -14,6 +14,7 @@ export interface TableAnchorRowData extends BaseTableRowData {
     href: string;
     hrefLabel: string;
     onRowClick?: (href: string) => void;
+    customRowClick?: () => void;
```
lfbergee: I'll second @Saegrov here. In general it's not a great pattern to have one prop invalidate the intent of another, as this will increase the complexity of the component and make it harder to use and understand.

This looks fine
```
...
 onRowClick={()=>console.log('the row were clicked')}
 customRowClick={()=>console.log('custom row')}
...
```

but onRowClick will never do anything..
```

eduardodir:
Delete this branch, the reason we contacted you yesterday was that it was the first thing we tried and it did not work, probably some issue with the open localhost that was tested.

PR #427, opened 2019-10-11
feat(scss): swap sass compiler to dart-sass

lfbergee:
## 📥 Proposed changes

dart-sass is the refrence implementation of sass, so it will get the latest features and fixes first

it is also pure javascript, node-sass is a wrapper of libsass, a c++ implementation, with its inherited problems of wrapping c++ node-sass is stil a bit faster than dart-sass, but the differenence for us is about 1 second for compiling all scss currently this enables us to utilize the new module system in scss, which should reduce could reduce css size and improve performance in production

the downside of this change is that webpack and parcel will by default not understand nested imports anymore. If that posibility is actually used by our clients is another case, I guess the only reason for doing that is that you want to complie the scss yourself.

| node-sass | dart-sass |
| --- | --- |
| 12.47s | 13.19s |
| 11.68s | 13.96s |
| 11.43s | 12.61s |
| 11.36s | 12.63s |
| 12.18s | 12.85s |
| Average |
| 11.824s | 12.992s |

## ☑️ Submission checklist

## 💬 Further comments

affects: @fremtind/jkl-accordion, @fremtind/jkl-bullet-list, @fremtind/jkl-button,
@fremtind/jkl-card-react, @fremtind/jkl-card, @fremtind/jkl-checkbox, @fremtind/jkl-datepicker,
@fremtind/jkl-divider-line, @fremtind/jkl-dropdown, @fremtind/jkl-field-group-react,
@fremtind/jkl-field-group, @fremtind/jkl-hamburger, @fremtind/jkl-loader, @fremtind/jkl-logo,
@fremtind/jkl-message-box, @fremtind/jkl-radio-button, @fremtind/jkl-table-react,
@fremtind/jkl-table, @fremtind/jkl-text-input, @fremtind/jkl-toggle-switch


gulpfile.js
```diff
@@ -1,12 +1,12 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
 const sass = require("gulp-sass");
-const importer = require("node-sass-tilde-importer");
 const postcss = require("gulp-postcss");
 const cssnano = require("cssnano");
 const autoprefixer = require("autoprefixer");
 const rename = require("gulp-rename");
+const Fiber = require("fibers");
```
lfbergee: Fibers is a libary made to bridge the gap for code written in a  synchronous language and executen in a asynchronous language. This allows us to compile asynchronously, while keeping more or less the performance.

Saegrov:
If you want to remove `node-sass` as a dependency, you must remove it from portal as well.

piofinn:
> Webstorm/IntelliJ developer experience suffers as Sass-plugin by Jetbrains plugin doesn't support adding load paths and this is the result:

This seems like it might be a problem. I know a lot of our developers use IntelliJ, and conditioning people to ignore errors feels like a bad idea 😅

lfbergee:
This comes from previous rather bad experiences with node-sass, however node-sass has been quite well behaved in Jøkul.

But I did make it work with the custom importer, so we can keep the tilde. It's not clear what the sass team will do, but it does seem like the ~ is a convention that will stay. 

Then will keep the current behaviour and get the benefit of using the canonical sass implementation and a native javascript library.

packages/table/table.scss
```diff
@@ -52,6 +52,7 @@ $highlight-color: $helhvit;
         }
 
         &--data-row {
+            -webkit-transition: background-color 150ms ease, box-shadow 150ms ease;
```
piofinn: This should be handled by autoprefixer?

lfbergee: yeah, I think I ran autoprefixer on sass files while testing, and this slipped through

PR #440, opened 2019-10-11
feat(dropdown): make dropdowns responsive

### LINKED ISSUE #424: Datepicker is only partially responsive
**Describe the bug**
Datepicker uses `text-input` for both the main input and the year input in the extended version, as well as a `dropdown` for the month chooser in the extended version. Right now the `text-input` has responsivity implemented, but the `dropdown` has not, leading to some wanky design. 

**To Reproduce**
1. Go to the datepicker in the portal
2. set `extended` as a prop on one of the examples
3. open the example in question
4. resize the window to something smaller than 768px
5. see error

**Expected behavior**
The `dropdown` should also be responsive

**Screenshots**
Skjermbilde 2019-10-09 kl  14 00 26

**Suggestion**
Implement responsivity in all remaining components, starting with `dropdown`

piofinn:
affects: @fremtind/jkl-dropdown-react, @fremtind/jkl-dropdown

## 📥 Proposed changes

* Make dropdowns scale responsively to screen size as per sketches.
* Add option to force compact mode

Fixes #424 

## ☑️ Submission checklist

## 💬 Further comments

### DIFF
--- a/packages/dropdown-react/src/Select.tsx
+++ b/packages/dropdown-react/src/Select.tsx
@@ -17,6 +17,7 @@ interface Props {
     variant?: LabelVariant;
     placeholder?: string;
     value?: string;
+    forceCompact?: boolean;
 }
 
 export function Select({
@@ -30,6 +31,7 @@ export function Select({
     variant,
     placeholder,
     value,
+    forceCompact,
     ...rest
 }: Props) {
     // If no value is given, set it to first item, or to empty string if there is a placeholder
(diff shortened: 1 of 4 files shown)


packages/dropdown-react/src/Select.tsx
```diff
@@ -45,14 +47,18 @@ export function Select({
...
-    const labelClassName = variant ? ` jkl-label--${variant}` : "";
+    const labelClassName = "jkl-label".concat(
+        variant ? ` jkl-label--${variant}` : "",
+        forceCompact ? ` jkl-label--compact` : "",
+    );
     return (
-        <label className={componentClassName}>
-            <span className={"jkl-label" + labelClassName}>{label}</span>
+        <label data-testid="jkl-dropdown" className={componentClassName}>
+            <span className={labelClassName}>{label}</span>
```
piofinn: This stuff gets repeated in a lot of components now. It feels like we need a separate`Label` component, maybe?

piofinn: This gets fixed in #442!

PR #443, opened 2019-10-16
Add deploy action

piofinn:
## 📥 Proposed changes

Add a deploy action using GitHub Actions, to deploy the portal to GitHub Pages on push to master.

## ☑️ Submission checklist

## 💬 Further comments

I'm using a third-party action here, but we can look into writing it from scratch if there are security concerns.


.github/workflows/test.yml
```diff
@@ -1,17 +1,19 @@
 name: CI
 
-on: [push]
+on:
+    push:
+        branches:
+            - !master
```
piofinn: There should be no need to run this action on `master` since everything should be built and tested before merge anyway. Alternatively we can keep this, and remove the build step from the deploy action.

PR #458, opened 2019-10-18
Rowdata nodearray extention

BjornBakke:
## 📥 Proposed changes

image
image

## ☑️ Submission checklist

### DIFF
--- a/packages/table-react/example/index.tsx
+++ b/packages/table-react/example/index.tsx
@@ -10,8 +10,8 @@ import "./index.scss";
 initTabListener();
 
 const columns = ["Dato", "Saksnummer", "Kundenummer", "Kundenavn", "Sakstype", "Milepæl", "Følger saken"];
-
-const rows: Array<string[] | TableAnchorRowData> = [
+const htmlElement = <span className="elementStyle" />;
+const rows: Array<(string[] | React.ReactNode[]) | TableAnchorRowData> = [
     ["24.02.2020", "20-1234567", "010203 99887", "Ola Nordmann", "Etterlatte", "Opprettet", "Siri Saksbehandler"],
     {
         href: "https://google.com",
@@ -30,6 +30,7 @@ const rows: Array<string[] | TableAnchorRowData> = [
     ["13.04.2019", "20-8382811", "010203 99887", "Kari Nordkvinne", "Etterlatte", "Opprettet", "Siri Saksbehandler"],
 
     ["31.07.2017", "20-1111", "010203 99887", "Kari Nordkvinne", "Etterlatte", "Opprettet", "Per Persen"],
+    [htmlElement, htmlElement, htmlElement, htmlElement, htmlElement, htmlElement, htmlElement],
 ];
 
 const TableDemo = () => {
--- a/packages/table-react/src/Table.tsx
+++ b/packages/table-react/src/Table.tsx
@@ -1,11 +1,11 @@
 import React from "react";
 import { TableAnchorRowData, TableRow } from "./TableRow";
 
-export type TableRowData = string[] | TableAnchorRowData;
+export type TableRowData = (string[] | React.ReactNodeArray) | TableAnchorRowData;
 
 interface Props {
     columns: string[];
-    rows: Array<TableRowData>;
+    rows: Array<TableRowData | React.ReactNodeArray>;
     className?: string;
 }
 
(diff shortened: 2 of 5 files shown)


lfbergee:
@all-contributors please add @BjornBakke for code

packages/table-react/src/Table.tsx
```diff
@@ -1,7 +1,7 @@
 import React from "react";
 import { TableAnchorRowData, TableRow } from "./TableRow";
 
-export type TableRowData = string[] | TableAnchorRowData;
+export type TableRowData = React.ReactNodeArray | TableAnchorRowData;
```
lfbergee: We usually destruct in import, so `import React, { ReactNodeArray }`, nitpicky

packages/table-react/example/index.tsx
```diff
@@ -10,8 +10,8 @@ import "./index.scss";
 initTabListener();
 
 const columns = ["Dato", "Saksnummer", "Kundenummer", "Kundenavn", "Sakstype", "Milepæl", "Følger saken"];
-
-const rows: Array<string[] | TableAnchorRowData> = [
+const htmlElement = <span className="elementStyle" />;
+const rows: Array<(string[] | React.ReactNode[]) | TableAnchorRowData> = [
```
piofinn: An array of arrays? 🤔 Should probably be
```jsx
const rows: Array<string | ReactNode | TableAnchorRowData>
```

lfbergee:
Also, really nitpicky, commit messages should be in english, we genereate the changelog from the commit messages, so it should be consistent.

lfbergee:
I do like the placeholder thingy, I think that can be a feature implemented in the base-table component, so then we don't get a bunch of different placeholder-implementations in the clients. Maybe we can event animate it a bit and make it look cool af

piofinn:
> I do like the placeholder thingy

Yes! I think we should aim to have skeleton versions of most types of content, since we have (too) many solutions that take a while to load data.

piofinn:
I couldn't resist making a little POC of an animated placeholder 😅 Heavily inspired by Carbon design skeleton elements here, but I think it works quite well!

```scss
.elementStyle {
    display: inline-block;
    position: relative;
    overflow: hidden;
    width: 100%;
    height: 20px;
    background-color: $himmelgra;

    &:before {
        content: "";
        width: 0%;
        height: 100%;
        position: absolute;
        top: 0;
        left: 0;
        background-color: $lys-gra;
        animation: load-animation 3500ms infinite ease-in-out;
    }
}

@keyframes load-animation {
    0% {
        width: 0%;
        left: 0;
        right: auto;
        opacity: 0.5;
    }

    25% {
        width: 100%;
        opacity: 0.4;
        left: 0;
        right: auto;
    }
    30% {
        width: 100%;
        left: auto;
        right: 0;
    }
    49% {
        width: 0%;

        left: auto;
        right: 0;
    }
    50% {
        width: 0%;
        left: 0;
        right: auto;
        opacity: 0.2;
    }

    70% {
        width: 100%;

        left: 0;
        right: auto;
    }
    75% {
        width: 100%;

        left: auto;
        right: 0;
    }
    100% {
        width: 0%;
        opacity: 0.4;
        left: auto;
        right: 0;
    }
}
```

Copy and paste into the `table-react` example stylesheet to preview 😊

Saegrov:
> I agree with @lfbergee about this maybe opening up the compnent too much.. Maybe we're heading towards delivering several `SomethingTable` components based on a base `Table` component? In that case, the base should probably accept `ReactNode`s, but allowing _all_ content into our only `Table` seems wrong somehow 😅
> 
> If there is a specific problem behind this PR, it might be possible to narrow down the allowed elements using the required `HTML**Element` types instead of `ReactNode`? **Edit:** example below
> 
> ```tsx
> type AllowedElement = HTMLButtonElement | HTMLSpanElement; // whatever is needed
> interface Props {
>     data: Array<string | AllowedElement>
>     ...
> }
> ```

This isn't possible with Typescript:

> By default the result of a JSX expression is typed as any. You can customize the type by specifying the JSX.Element interface. However, it is not possible to retrieve type information about the element, attributes or children of the JSX from this interface. It is a black box.

https://www.typescriptlang.org/docs/handbook/jsx.html#the-jsx-result-type

piofinn:
Etter å ha snakket med @BjornBakke har vi valgt å kulle denne PR'en og heller jobbe videre med å finne en løsning sentralt, da de er litt presset for tid på Flyt-teamet. Vi lager et issue på skeleton-/loading-versjoner av komponenter generelt, og et for tabell spesifikt (siden det er et reelt behov).

PR #460, opened 2019-10-18
feat(button): implement forceCompact on buttons

piofinn:
## 📥 Proposed changes

Implement ability to force compact mode on buttons.

## ☑️ Submission checklist

## 💬 Further comments

affects: @fremtind/jkl-button-react, @fremtind/jkl-button

### DIFF
--- a/packages/button-react/example/index.tsx
+++ b/packages/button-react/example/index.tsx
@@ -1,8 +1,11 @@
 import React from "react";
 import ReactDOM from "react-dom";
 import { PrimaryButton, SecondaryButton, TertiaryButton } from "../src";
+import { initTabListener } from "@fremtind/jkl-core";
 import "@fremtind/jkl-button/button.css";
-import "@fremtind/jkl-core/core.min.css";
+import "@fremtind/jkl-core/core.css";
+
+initTabListener();
 
 function onClick() {
     console.log("Hello!");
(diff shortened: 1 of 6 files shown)


packages/button-react/example/index.tsx
```diff
@@ -1,23 +1,43 @@
...
+                Svar senere
+            </SecondaryButton>
         </div>
-        <div style={{ margin: "20px" }}>
-            <TertiaryButton onClick={onClick}>Tertiary</TertiaryButton>
+        <div className="side-by-side">
+            <TertiaryButton onClick={onClick}>Avbryt</TertiaryButton>
+            <TertiaryButton forceCompact onClick={onClick}>
+                Avbryt
+            </TertiaryButton>
```
lfbergee: If you view this on a mobile-device, it looks the same. That might not be an issue, but the demo kinda don't work in that case.

Is there any case, besides a demo like this to force into large?

piofinn: Yeah, I see your point! It does still demonstrate the functionality, but it's impossible to see the difference 🤔 It looks like we're heading towards keeping the big size on mobile, so I see no point in implementing a `forceLarge` functionality per now…

packages/button-react/src/Button.test.tsx
```diff
@@ -1,11 +1,29 @@
...
+            </>,
+        );
+
+        const primaryButton = getByText("primary");
+        const secondaryButton = getByText("secondary");
+        const tertiaryButton = getByText("tertiary");
+
+        expect(primaryButton).toHaveClass("jkl-button--primary");
+        expect(secondaryButton).toHaveClass("jkl-button--secondary");
+        expect(tertiaryButton).toHaveClass("jkl-button--tertiary");
```
lfbergee: I don't like multiple test in a test.

I'd prefer to split this into three test. Then each will report as its own test.

```
[
        { name: "Primary", component: PrimaryButton },
        { name: "Secondary", component: SecondaryButton },
        { name: "Tertiary", component: TertiaryButton },
    ].map((buttonVariant) => {
        it(`should work for ${buttonVariant}`, () => {
            const C = buttonVariant.component;

            const { getByText } = render(<C onClick={() => {}}>{buttonVariant.name}</C>);
            expect(getByText(buttonVariant.name)).toHaveClass(`jkl-button--${buttonVariant.name.toLowerCase()}`);
        });
    });
```

piofinn: Ooh, I like this!

packages/button-react/example/index.tsx
lfbergee: For sure, we should not implement forceLarge just for test/demo purposes.

Issue #426, opened 2019-10-11
Input fields /w suffix + inline input fields


Mikkelinski:
We might need the ability to add a suffix to our input fields (e.g. _______ years / ______ kroner) or place input fields inline where the surrounding context replaces the label.

How should it look / act? Using the existing input field as a baseline

vegardsoerlie:
A bit eager on the close button there 😅 

Before closing we need a better way to render a proper suffix input field. As it is now we can use `InlineTextField` but it's a bit hacky and already has its own `Label` that's only visible for screen readers.

piofinn:
We're closing this for the time being, since the inline version covers current needs. We'll open a new issue if the need for suffixes surfaces again

PR #462, opened 2019-10-18
feat(text-field): implement in line version of TextField

piofinn:
## 📥 Proposed changes

Implement inline (in-text) version of `TextField` as discussed in #426 . This also adds a screenreader-only version of the label so it can be hidden without creating an empty `<label>` element.

## ☑️ Submission checklist

## 💬 Further comments

affects: @fremtind/jkl-core, @fremtind/jkl-text-input-react, @fremtind/jkl-text-input,
@fremtind/jkl-typography-react

### Implementation details

- Repurposes the existing `inline` prop (which had no real effect) for this, which strictly speaking breaks current functionality; We should give a heads-up to the teams before publishing.
- Adds both a manual and an automatic way of controlling input length: The new `width` prop takes a string, which is passed to the css `width` property as-is; If no `width` is passed, the field will use any `maxLength` value of less than 15 to decide the input length. This system ensures good affordance, and enables fine control.

### Further implications

We should probably create similar versions of the `Dropdown`/`Select` and `DatePicker` components.

### DIFF
--- a/packages/text-input-react/src/TextField.tsx
+++ b/packages/text-input-react/src/TextField.tsx
@@ -1,4 +1,4 @@
-import React, { ChangeEvent, FocusEvent } from "react";
+import React, { ChangeEvent, FocusEvent, CSSProperties } from "react";
 import { LabelVariant } from "@fremtind/jkl-core";
 import { Label, SupportLabel } from "@fremtind/jkl-typography-react";
 
@@ -46,13 +46,13 @@ export const TextField = ({
         forceCompact ? " jkl-text-field--compact" : "",
         className ? ` ${className}` : "",
     );
-    function getLength(): string | undefined {
+    function getWidthAsStyle(): CSSProperties | undefined {
         if (width) {
-            return width; // prioritize width
+            return { width }; // prioritize width prop
         }
 
         if (maxLength && maxLength < 15) {
-            return `${maxLength + 3}ch`; // else use maxLength if not large
+            return { width: `${maxLength + 3}ch` }; // else use maxLength if not large
         }
 
         return undefined;
@@ -71,7 +71,7 @@ export const TextField = ({
                 readOnly={readOnly}
                 value={value}
                 maxLength={maxLength}
-                style={getLength() ? { width: getLength() } : undefined}
+                style={getWidthAsStyle()}
                 {...rest}
             />
             {!inline && <SupportLabel helpLabel={helpLabel} errorLabel={errorLabel} forceCompact={forceCompact} />}
--- a/packages/text-input-react/src/TextField.tsx
+++ b/packages/text-input-react/src/TextField.tsx
@@ -21,6 +21,7 @@ interface Props {
     variant?: LabelVariant;
     forceCompact?: boolean;
     maxLength?: number;
+    width?: string;
 }
 
 export const TextField = ({
@@ -37,6 +38,7 @@ export const TextField = ({
     variant,
     forceCompact,
     maxLength,
+    width,
     ...rest
 }: Props) => {
     const componentClassName = "jkl-text-field".concat(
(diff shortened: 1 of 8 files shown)


packages/text-input-react/src/TextField.tsx
```diff
@@ -37,16 +38,26 @@ export const TextField = ({
...
+    width,
     ...rest
 }: Props) => {
     const componentClassName = "jkl-text-field".concat(
         inline ? " jkl-text-field--inline" : "",
         forceCompact ? " jkl-text-field--compact" : "",
         className ? ` ${className}` : "",
     );
+    function getLength(): string | undefined {
+        if (width) return width; // prioritize width
```
lfbergee: No no, this should be a lint-rule. It sort of looks pretty, but it degrades so fast, please add `{}`

packages/text-input-react/src/TextField.tsx
```diff
@@ -58,9 +69,10 @@ export const TextField = ({
                 readOnly={readOnly}
                 value={value}
                 maxLength={maxLength}
+                style={getLength() ? { width: getLength() } : undefined}
```
lfbergee: Maybe prettier if getLength returns the object, then we'll reduce the need of this turnary.

Not a huge fan of inlinestyle, but might not be a great way to get around it here.

packages/text-input/text-input.scss
```diff
@@ -23,7 +23,14 @@ $textarea-expanded-height--small-device: $input-height--small-device * 7 + $bott
 
     &--inline {
         display: inline-block;
-        vertical-align: bottom;
+        width: unset;
+        margin-top: rem(-4px);
+        vertical-align: top;
+
+        .jkl-p & .jkl-text-field__input {
+            padding: 0; // $component-spacing--small;
```
lfbergee: Intentional?

packages/text-input-react/src/TextField.tsx
```diff
@@ -58,9 +69,10 @@ export const TextField = ({
...
                 {...rest}
             />
-            <SupportLabel helpLabel={helpLabel} errorLabel={errorLabel} forceCompact={forceCompact} />
+            {!inline && <SupportLabel helpLabel={helpLabel} errorLabel={errorLabel} forceCompact={forceCompact} />}
```
lfbergee: Do we want inline to be it's own component?

I think we shouldn't have props that kills other props without warning, so at least if we do this we should throw a warning or error while in DEV that inline and help/error do not match

packages/text-input-react/src/TextField.tsx
lfbergee: Sortof the same goes with width killing max-width. In css is the other way around, so we should document that behaviour and again maybe throw a warning/error.

packages/text-input-react/src/TextField.tsx
piofinn: Haha, sorry 😅

packages/text-input/text-input.scss
piofinn: Nope!

packages/text-input-react/src/TextField.tsx
piofinn: Yeah, that is a possibility to consider. In that case we might want to move the proposed `ActionTextField` into this package as well, rather than have three separate pkgs...

packages/text-input-react/src/TextField.tsx
piofinn: I like the idea of returning the style object. I don't particularly like ternaries in props either, just didn't think that far.

I'm not sure exactly what you mean by killing `max-width`? That the inline will override it in specificity? In that case I feel like the intentionality in specifying a width in the prop is enough explanation..

lfbergee: Yeah, my comment did not make that much sense. So, just disregard that.

lfbergee:
Still think we should avoid or at lease warn developers about props that override each other

piofinn:
> Still think we should avoid or at lease warn developers about props that override each other

Yes, I agree. Will push a commit with a warning in dev now, and make a test branch with the inline version as a separate component!

packages/text-input-react/src/TextField.tsx
vegardsoerlie: I support the idea of extracting inline as a component and including ActionTextField (and ActionIcon?) in this package as well.

piofinn: As long as we're keeping to just a handful of icons, I agree that thay can live here until we use them in another input type (e.g. select) as well.

Issue #464, opened 2019-10-22
Set fallback font behavior


lfbergee:
`font-display: fallback`

Set to normalize browser behavior when the font is not yet loaded. Fallback seems to be the best compromise between user experience and our desire to use our beautiful font

piofinn:
Seems like a good idea!

PR #474, opened 2019-10-23
feat(core): specify font fallback strategy

piofinn:
## 📥 Proposed changes

explicitly set font fallback strategy to secure consistency across browsers

## ☑️ Submission checklist

## 💬 Further comments

affects: @fremtind/jkl-core

ISSUES CLOSED: #464

### DIFF
--- a/packages/core/font.scss
+++ b/packages/core/font.scss
@@ -7,6 +7,7 @@
             format("woff");
     font-style: normal;
     font-weight: normal;
+    font-display: fallback;
 }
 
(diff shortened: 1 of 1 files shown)


lfbergee:
For fremtinden:
fallback: Acts as a compromise between the auto and swap values. The browser will hide the text for about 100ms and, if the font has not yet been downloaded, will use the fallback text. It will swap to the new font after it is downloaded, but only during a short swap period (probably 3 seconds).

PR #456, opened 2019-10-17
Icon button

frisol97:
## 📥 Proposed changes

There is a need for text-field with icons, so this is a suggestion of how it could be implemented. An action-icon acts like a button and should be used when it is necessary to make the use understand that this input-field (for example) is a search-field. 

This pull request adds action-icon and action-text-field components.

## ☑️ Submission checklist

-   [x] I have read the [CONTRIBUTING](https://github.com/fremtind/jokul/blob/master/CONTRIBUTING.md) doc
-   [x] `yarn build` works locally with my changes
-   [x] I have added tests that prove my fix is effective or that my feature works
-   [x] I have added necessary documentation (if appropriate)

## 💬 Further comments

A better solution is needed.
1) We need a proposal from designer on how these icons should look.
2) The icons should not be a component themselves. Could we move it to the action-text-field or core?

closes #448

PR #486, opened 2019-10-29
Text input merge

vegardsoerlie:
## 📥 Proposed changes

Merge the four text-input components 'TextField', 'TextArea', 'TextField' variant 'inline' and 'ActionTextField' into the same package (text-input-react). Additionally the 'TextField' variant 'inline' is now refactored as its own component called 'InlineTextField'.

The icons and ActionIcon components are also part of the 'text-input-react' package as of now, but this is subject to change if we see a need for icons elsewhere.

## ☑️ Submission checklist

## 💬 Further comments

This basically replaces pull request #456 as all proposed features added there are included here.

I feel TextField, TextArea and InlineTextField are appropriately named, but ActionTextField could need a clearer name.

ActionTextField is in need of a functioning compact mode, this is not included yet as we await the design of the component. Also, when errorLabel is used on this component, the div size increases and the icon is no longer correctly positioned.


piofinn:
Lurer på om denne bør lages "på nytt" siden det har blitt med så mange commits fra master her… Siden det egentlig bare er én commit ville jeg bare kopiert hele `text-field` og `text-field-react`-mappene fra denne branchen, og limt dem inn i en ny branch fra nyeste versjon av `icon-button`-branchen. Da _bør_ du få en ren commit med bare endringene du har gjort.

PR #519, opened 2019-11-20
fix(select-react): use valuePair label as initial display value

### LINKED ISSUE #517: Initial display value for select with value pair should use label instead of value
**Describe the bug**

Initial render of select with `ValuePair` shows the value instead of the label. I.e. `{ value: 9, label: 'My pretty little string' }` will display `9` to the user where one would expect `'My pretty little string'`.

This is likely resolved as a side-effect in https://github.com/fremtind/jokul/commit/1fa7c2f4a1046be27298d73209b43b8e6cbcb341 which is not yet released. In any case, it would be best to should account for `ValuePair` explicitly.

piofinn:
affects: @fremtind/jkl-select-react

## 📥 Proposed changes

Be sure to use the `label` from a `ValuePair` as the initial displayed value in the `Select` component, instead of the `value`.

## ☑️ Submission checklist

## 💬 Further comments

Fixes #517


piofinn:
Innså at fix-en min ikke fiksa noe, så måtte gjøre det på nytt! 😅

PR #673, opened 2020-01-14
docs(portal): content for cards Profilelementer and Utvikler

eduardodir:
Added two new texts


piofinn:
Do these need to be their own files? This seems like it could just be hardcoded until we potentially get a CMS, as they probably won't be changing much..

The texts are fine!

Gorild:
@piofinn for now we prefer to have all texts in md files an very little hard coded texts

lfbergee:
@all-contributors please add @eduardodir for code

Issue #745, opened 2020-02-04
Simplify use of classNames internally in components


joms:
Currently, a repeating pattern for using `classNames` is using string concatenation. It's a bit hard to read, not really intuitive, and feels repetitive to generate.

I'm not sure what the guidelines for using external libraries are, but one could use the [classnames package](https://www.npmjs.com/package/classnames), or create a simplified utility function for generating classNames.

This would change

```
const componentClassName = "jkl-accordion-item".concat(
    isOpen ? " jkl-accordion-item--expanded" : "",
    className ? ` ${className}` : "",
);
```

into

```
const componentClassName = className(
	"jkl-accordion-item", {
	    "jkl-accordion-item--expanded": isOpen,
    },
	className
);

// alternatively
const componentClassName = className({
		"jkl-accordion-item",
	    "jkl-accordion-item--expanded": isOpen,
    },
	className
);
```

Pros of the proposed way of doing this:
- Clear definition which is easy to read and get an overview of what's happening and how each class is controlled
- No need for hard to read ternary expressions
- Easy to control dynamic behaviour within the component
- `className` prop could be a dynamic object controlling styles from outside the component

piofinn:
The current pattern stems from when we had few dynamic classes in the library. We considered `classnames` back then, but decided to wait.

This is probably as good a time as any to start using it, so I'm all for it! Sounds like a great little refactor if you want to get to know the components better, @joms 😊

joms:
Awesome @piofinn! Would we benefit for creating a `jkl-util` package where we could store our utilities (I assume that this won't be the last one), or are there other patterns for creating utility functionality?

piofinn:
We have discussed a `jkl-util` package before, and it might be a good idea at some point. We already have a package with React hooks that could be moved there if we find the need.

For this case though, I think it's easiest to just use the `classnames` package. It works well and seems to have types available. We don't have any formal rules about new packages, but the practice has been "add it if you think it adds value, and we'll discuss it in the PR" 😊

joms:
Refactored quite a lot of classname definitions to use the classnames package. Will look for more definitions to replace tomorrow before starting CR.

joms:
🎉

PR #750, opened 2020-02-05
Implement classnames package to simplify how classes for elements are generated and controlled

joms:
## 📥 Proposed changes

To simplify how generated classnames for elements are generated and controlled, I in #745 proposed to introduce the [classnames](https://www.npmjs.com/package/classnames) package, or create a similar solution.

I have introduced the package, as it's a fairly small package where the whole package is utilised. This allows us to define classname strings in a more nimble and intuitive way.

## ☑️ Submission checklist

-   [X] I have read the [CONTRIBUTING](https://github.com/fremtind/jokul/blob/master/CONTRIBUTING.md) doc
-   [X] `yarn build` works locally with my changes
-   [X] I have added tests that prove my fix is effective or that my feature works
-   [X] I have added necessary documentation (if appropriate)

## 💬 Further comments

Due to the nature of how the package is exported, I had to define a named export in the rollup configuration (see 9d01fe00edc80ad6a8e6412dcb63c8dbc79341a5). Not having this definition caused the building of some packages to fail, even though others worked just fine. Not sure why this occurred. This is confirmed by both the testing pipeline in GitHub Actions, and trying to build locally.

If someone feels like this the wrong way to solve this issue, please approach me so we can do a debugging session🚀


packages/accordion-react/src/AccordionItem.tsx
```diff
@@ -4,6 +4,7 @@ Hopefully someone (us?) will write types for it sometime soon */
 import CoreToggle from "@nrk/core-toggle/jsx";
 import React, { ReactNode, useState } from "react";
 import { useAnimatedHeight } from "@fremtind/jkl-react-hooks";
+import cx from "classnames";
```
piofinn: Tror jeg ville foretrukket `classNames` eller tilsvarende som navn på importen. Det fjerner enhver tvil om hva som foregår når funksjonen kalles 😊

package.json
```diff
@@ -18,6 +18,8 @@
         "release": "lerna publish"
     },
     "dependencies": {
+        "@types/classnames": "^2.2.9",
```
piofinn: Dette kan vel være en devDependency? Ser for øvrig ut som `classnames` har typedefinisjoner, [men at de ikke blir eksportert med pakken ](https://github.com/JedWatson/classnames/issues/209)😑

packages/accordion-react/src/AccordionItem.tsx
joms: Enig! `cx` satt i fingrene fra tidligere bruk 🙃

PR #785, opened 2020-02-19
feat(text-input): tweak action icon placement and add focus state

piofinn:
affects: @fremtind/jkl-text-input-react, @fremtind/jkl-text-input

## 📥 Proposed changes

Tweaks size and position of the action icon in `ActionTextField`, and adds keyboard focus indication to it in the form of blue focus color.

## ☑️ Submission checklist

## 💬 Further comments

We should probably find a better way of indicating focus state, but this is better than nothing, and was easy to implement via CSS.

### DIFF
--- a/packages/text-input-react/src/Icons/IconClear.tsx
+++ b/packages/text-input-react/src/Icons/IconClear.tsx
@@ -1,7 +1,8 @@
 import React from "react";
 
 export const IconClear = () => (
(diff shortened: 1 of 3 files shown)


piofinn:
> Det ser ut som om søkeikonet har blitt kuttet av med noen piksler helt i toppen. Mulig det var sånn før også, men at det ikke ble catchet

Godt sett! Kom nok litt utenfor viewbox da jeg endret `stroke-width` til `1.5`. Funka fint når jeg flytta bittelitt på sirkelen 😊

PR #812, opened 2020-03-06
feat(core): add support for prefers-reduced-motion

piofinn:
## 📥 Proposed changes

Turn off all transitions and animations when the user prefers reduced motion

## ☑️ Submission checklist

## 💬 Further comments

affects: @fremtind/jkl-core


piofinn:
> Later on

We also need to define a strategy for javascript animations. I don't think we can express our visual identity properly without them, and they will probably be most relevant for `prefers-reduced-motion`.

PR #817, opened 2020-03-06
fix(react-hooks): correct return signature in useAnimatedHeight

### LINKED ISSUE #768: Wrong Typescript type for useAnimatedHeight
**Describe the bug**
The typescript type for the return type of `useAnimatedHeight` is wrong.

The following code produces two different typescript compile errors.
The code runs absolutely fine runtime but produces a typescript compile error.
```
  const [ref] = useAnimatedHeight<HTMLElement>(erEkspandert);
  
  if(ref.current) {    
    
  }
```

With the current implementation of `useAnimatedHeight` this is the same as this code:

```
  const ref1 = useRef<HTMLElement>();
  
  if(ref1.current) {
    // No error 
  }
```

**Errors** 
1. `TS2531: Object is possible null`. 
2. `TS2339: Property 'current' does not exist on type '((instance: HTMLElement | null) => void) | RefObject<HTMLElement>'.`

**To Reproduce**
See full example here:
https://codesandbox.io/s/wonderful-hopper-6wduy?fontsize=14&hidenavigation=1&theme=dark

**Expected behavior**
Code should not produce any typescript errors.

piofinn:
📥 Proposed changes

Correct the return signature of `useAnimatedHeight` to return a `RefObject` instead of a `Ref`.

## ☑️ Submission checklist

## 💬 Further comments

affects: @fremtind/jkl-react-hooks
Closes #768


piofinn:
> `MutableRefObject<T>` is the correct return type.

Are you sure? `useRef<T>(null)` (which is used in the hook) returns a `RefObject<T>`.

Hovewer `useRef(null)`, `useRef<T>(someT)` or `useRef<T | null>(null)` will return a `MutableRefObject<T>`.

In any case, your example works fine for me with the new return value:
Skjermbilde 2020-03-09 kl  10 32 29
Skjermbilde 2020-03-09 kl  10 53 11

I might just have misunderstood the types here, though 😅

Saegrov:
> > `MutableRefObject<T>` is the correct return type.
> 
> Are you sure? `useRef<T>(null)` (which is used in the hook) returns a `RefObject<T>`.

You are right! Your change does work. It must have been a hangup with my IDE because it was showing typescript-errors.

PR #860, opened 2020-03-24
Fix React hooks and add linting

piofinn:
## 📥 Proposed changes

Fixes some unsafe or possibly buggy uses of hooks, and adds linting rules for hooks.

## ☑️ Submission checklist

### DIFF
--- a/packages/text-input-react/src/TextArea.test.tsx
+++ b/packages/text-input-react/src/TextArea.test.tsx
@@ -1,5 +1,5 @@
-import React, { useState } from "react";
-import { cleanup, render, fireEvent, act } from "@testing-library/react";
+import React from "react";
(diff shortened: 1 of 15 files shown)


packages/text-input-react/src/TextArea.test.tsx
```diff
@@ -133,4 +133,5 @@ describe("a11y", () => {
     });
 });
 
+/* eslint-disable */
 type PropTypeFor<T extends (arg: any) => any> = T extends (arg: infer P) => any ? P : never;
```
piofinn: @Saegrov kanskje du vil kommentere hvorfor denne må være som den er?

Saegrov: Kan det være på grunn av at vi bruker `any`?

PR #861, opened 2020-03-24
feat: update typescript, prettier and babel

lfbergee:
## 📥 Proposed changes

typescript 3.8, nessestiate an upgrade of babel, prettier and eslint to use new features. Prettier is a major upgrade, but almost no changes to our project, as prettier has aligned their defaults pretty much with our settings.

The somewhat random changes here is due to the upgrade of prettier 2. To avoid getting these updates in an unreleated PR, `eslint --fix` has been executed on the entire project.

Also, I think its about time we say that Jokul isnt 0.0.0, so howabout 1.0.0?

## ☑️ Submission checklist

## 💬 Further comments

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html
https://prettier.io/blog/2020/03/21/2.0.0.html

### DIFF
--- a/packages/table-react/src/TableData.tsx
+++ b/packages/table-react/src/TableData.tsx
@@ -26,9 +26,7 @@ export function TableData({ data, row, isFirstCell }: Props) {
                 >
                     {row.hrefLabel}
                 </a>
-            ) : (
-                undefined
-            )}
+            ) : undefined}
             {data}
         </td>
     );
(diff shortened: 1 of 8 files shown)


packages/table-react/src/TableData.tsx
```diff
@@ -26,9 +26,7 @@ export function TableData({ data, row, isFirstCell }: Props) {
                 >
                     {row.hrefLabel}
                 </a>
-            ) : (
-                undefined
-            )}
+            ) : undefined}
```
lfbergee: Good job prettier 2.0, this is much better

PR #891, opened 2020-04-07
fix(codeblock): account for code blocks without language

piofinn:
## 📥 Proposed changes

Fixes a bug in the `CodeBlock` typographic component in the Portal, that caused page builds to fail if a Markdown code block had no assigned language.

## ☑️ Submission checklist

## 💬 Further comments

affects: @fremtind/portal

### DIFF
--- a/portal/src/components/Typography/Typography.tsx
+++ b/portal/src/components/Typography/Typography.tsx
@@ -56,7 +56,7 @@ export const CodeBlock: React.FC = ({ children, ...rest }) => {
         // if there is more than one child, or it is text return a normal pre
         return <pre {...rest}>{children}</pre>;
     }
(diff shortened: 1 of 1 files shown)


portal/src/components/Typography/Typography.tsx
```diff
@@ -56,7 +56,7 @@ export const CodeBlock: React.FC = ({ children, ...rest }) => {
         // if there is more than one child, or it is text return a normal pre
         return <pre {...rest}>{children}</pre>;
     }
-    const language = child.props.className.replace("language-", "");
+    const language = child.props.className?.replace("language-", "");
```
lfbergee, line 59: optional chaining ftw

PR #921, opened 2020-04-24
fix(example-knobs): use unique field names on knobs

piofinn:
## 📥 Proposed changes

Use unique field names in the `ComponentExample` component, to avoid issues when several examples with the same variable names are used on the same page.

## ☑️ Submission checklist

## 💬 Further comments

ISSUES CLOSED: #920

affects: @fremtind/jkl-portal-components

### DIFF
--- a/packages/portal-components/package.json
+++ b/packages/portal-components/package.json
@@ -56,7 +56,8 @@
         "url": "https://github.com/fremtind/jokul/issues"
     },
     "dependencies": {
-        "@fremtind/jkl-radio-button-react": "^1.3.5"
+        "classnames": "^2.2.6",
+        "nanoid": "^3.1.3"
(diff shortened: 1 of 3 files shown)


packages/portal-components/package.json
```diff
@@ -56,7 +56,8 @@
         "url": "https://github.com/fremtind/jokul/issues"
     },
     "dependencies": {
-        "@fremtind/jkl-radio-button-react": "^1.3.5"
+        "classnames": "^2.2.6",
```
lfbergee, line 59: Why this?

piofinn: It was used by the component without being a dependency, so I added it. Should probably have been it's own commit 😅

PR #1435, opened 2020-10-05
refactor(hooks): use optional chaining

deanshub:
## 📥 Proposed changes

A more readable code since you're using a newer typescript version.

## ☑️ Submission checklist


lfbergee (changes requested):
Hi,
thanks for the contribution. The change is fine, but we need the commit message to abide by the commitlint rules. So please reword to something like `refactor(hooks): use optional chaining`

We use the commit messages to generate the releases and changelogs, so its important for us. You can use `yarn commit` command to help you generate a valid message, just remeber to add the changes to stage first.

lfbergee (approved):
Thanks @deanshub

PR #1784, opened 2021-02-01
fix(button): fix rerender issue

lfbergee:
## 📥 Proposed changes
affects: @fremtind/jkl-button-react

ISSUES CLOSED:  #1783

defining button inside button caused an rerender all the way out to the DOM for each change in the
react tree

## ☑️ Submission checklist

## 💬 Further comments

### DIFF
--- a/packages/button-react/src/Button.tsx
+++ b/packages/button-react/src/Button.tsx
@@ -1,17 +1,8 @@
-import React, { forwardRef, ButtonHTMLAttributes, TouchEvent } from "react";
+import React, { forwardRef, TouchEvent } from "react";
 import classNames from "classnames";
 import { Loader } from "@fremtind/jkl-loader-react";
-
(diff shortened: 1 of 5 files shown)


packages/button-react/src/Button.tsx
```diff
@@ -1,17 +1,8 @@
-import React, { forwardRef, ButtonHTMLAttributes, TouchEvent } from "react";
+import React, { forwardRef, TouchEvent } from "react";
```
Saegrov, line 1: Skal ikke legge inn testen som er lagt inn i issue #1783 ?

Saegrov (approved):

packages/button-react/src/Button.tsx
lfbergee: Jo, det hadde vel kanskje vært smart

piofinn (approved):

PR #3735, opened 2023-10-26
chore: increase threshold for visual diff regression tests

ivarni:
What exactly this number should be will probably be a matter of trial and error but it's becoming clear that with the current workflow setup it can't be the default of 0.01

## 🎯 Sjekkliste

-   [x] `pnpm build` og `pnpm ci:test` gir ingen feil


ivarni:
oops, her var det noe som sneik seg med som ikke skulle det. Fikser.

PR #4181, opened 2024-10-17
chore: copies .d.ts files to where we expect them to be

ivarni:
The dts plugin has for unknown reasons suddenly decided to dump the generated files in a different subtree. This happens even with the exact same configuration and the exact same resolved version from npm so the only remaining explanation I can think of is that the amount of generated files prompted the plugin to change the output path. I can find no obvious way of telling it to do otherwise and I don't like this structure so for the time being this is a workaround

### DIFF
--- a/packages/jokul/vite.build.config.mjs
+++ b/packages/jokul/vite.build.config.mjs
@@ -1,4 +1,5 @@
 import { fileURLToPath } from "node:url";
+import { cpSync } from "node:fs";
 import { extname, relative, resolve } from "path";
 import terser from "@rollup/plugin-terser";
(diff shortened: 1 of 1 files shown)


ivarni:
Omskrev commit-kommentar fra `chore` til `fix`, ellers blir det jo ikke deployet noe nytt :)

kristianulv23 (approved):

BjorneOma (approved):
Kan ikkje sei at eg kjenner ting så godt at eg veit at dette er riktig måte, men funker det så funker det :)
Dette føles litt som ein ting som "berre skulle funka"?

PR #4399, opened 2025-01-02
chore: temporarily replace react-ink with gum

ivarni:


ivarni:
For å komme seg rundt https://github.com/vadimdemedes/ink/issues/688

